PR #21, opened 2023-04-19
Cleanup

atyoshioka:
- Change the return type of `get_columns_for_top_n` from `Vec<i32>` to `HashSet<i32>` since the returned value is immediately converted to `HashSet<i32>`
- Directly use `DEFAULT_NUMBER_OF_COLUMN ` in the `limited_top_n_of_clusters ` function

### DIFF
--- a/src/top_n.rs
+++ b/src/top_n.rs
@@ -237,7 +237,6 @@ fn total_of_top_n(
 fn limited_top_n_of_clusters(
     top_n_of_clusters: HashMap<i32, HashMap<usize, HashMap<String, i64>>>,
     limit_rate: f64,
-    default_number: usize,
 ) -> HashMap<usize, HashMap<String, i64>> {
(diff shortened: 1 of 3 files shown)


msk:
> * Change the data type of `DEFAULT_NUMBER_OF_COLUMN ` to `i64` and directly use it in the `limited_top_n_of_clusters ` function

IMO, this is better to be left `usize`. By its very nature, the number of columns can never be a negative value and in Rust the `usize` type is commonly used for this purpose. Although the constant must be cast to a signed integer for comparison with a signed variable, this is a separate concern. In such cases, it is recommended to perform the casting operation at the point of comparison, ensuring that the conversion is handled explicitly and safely. This way, we preserve the unsigned nature of the constant while still enabling comparisons with signed variables.

atyoshioka:
I'm not sure. I agree that if the data type is `usize`, it makes more sense, but `DEFAULT_NUMBER_OF_COLUMN` is a `const` variable, so it never becomes a negative value. Anyway I reverted the change.

msk (approved):

PR #22, opened 2023-04-20
Add `sensors`, `confidence`, and `learning_methods` in `EventFilter`

sehkone:
The `eventList` API requires the functionality to allow users to search for events based on sensors, confidence level, and learning methods (either unsupervised or semi-supervised). Therefore, the ability to filter events using these parameters is included as a feature.

### DIFF
--- a/src/migration.rs
+++ b/src/migration.rs
@@ -11,11 +11,12 @@ use bincode::Options;
 use semver::{Version, VersionReq};
 use serde::{Deserialize, Serialize};
 
-/// The version requirement for the database format to be compatible with the
-/// current version. When the database format changes, this requirement must be
-/// updated to match the new version, and the migration code must be added to
-/// the `migrate_data_dir` function.
-const DATABASE_VERSION_REQ: &str = "=0.5.0-alpha.4";
+/// The latest break version for the database format to be compatible with the
+/// current version. When the database format changes, this must be updated to
+/// match the new break version, and the migration code must be added to the
+/// `migrate_data_dir` function. This should be complete and include any alpha
+/// or beta version designations if applicable.
+const BREAK_VERSION: &str = "0.5.0-alpha.4";
 
 /// Migrates the data directory to the up-to-date format if necessary.
 ///
@@ -28,7 +29,7 @@ pub fn migrate_data_dir<P: AsRef<Path>>(data_dir: P, backup_dir: P) -> Result<()
     let data_dir = data_dir.as_ref();
     let backup_dir = backup_dir.as_ref();
 
-    let compatible = VersionReq::parse(DATABASE_VERSION_REQ).expect("valid version requirement");
+    let break_version = Version::parse(BREAK_VERSION).expect("valid version requirement");
 
     let (data, data_ver) = retrieve_or_create_version(data_dir)?;
     let (backup, backup_ver) = retrieve_or_create_version(backup_dir)?;
@@ -40,7 +41,7 @@ pub fn migrate_data_dir<P: AsRef<Path>>(data_dir: P, backup_dir: P) -> Result<()
     }
 
     let mut version = data_ver;
-    if compatible.matches(&version) {
+    if version >= break_version {
         // updates version on file to current
 
         create_version_file(&backup).context("failed to update VERSION")?;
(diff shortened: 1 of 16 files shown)


sehkone:
Since `cargo clippy` generates so many errors, I had to put the below in `lib.rs`.

`#![allow(clippy::missing_errors_doc)]`

msk:
> Since `cargo clippy` generates so many errors, I had to put the below in `lib.rs`.
> 
> `#![allow(clippy::missing_errors_doc)]`

We should avoid disabling clippy lints in the source codes. Please set this in your local clippy.toml instead.

sehkone:
> > Since `cargo clippy` generates so many errors, I had to put the below in `lib.rs`.
> > `#![allow(clippy::missing_errors_doc)]`
> 
> We should avoid disabling clippy lints in the source codes. Please set this in your local clippy.toml instead.

sehkone:
-

src/migration.rs
```diff
@@ -15,7 +15,7 @@ use serde::{Deserialize, Serialize};
 /// current version. When the database format changes, this requirement must be
 /// updated to match the new version, and the migration code must be added to
 /// the `migrate_data_dir` function.
-const DATABASE_VERSION_REQ: &str = ">=0.3.0,<0.5.0-alpha.4";
+const DATABASE_VERSION_REQ: &str = ">=0.3.0,<0.5.0-alpha.5";
```
msk, line 18: This is the version that doesn't need migration. Since this PR makes the previous version incompatible, this should be`"=0.5.0-alpha.5"`.

src/migration.rs
```diff
@@ -47,11 +47,24 @@ pub fn migrate_data_dir<P: AsRef<Path>>(data_dir: P, backup_dir: P) -> Result<()
         return create_version_file(&data).context("failed to update VERSION");
     }
 
-    let migration = vec![(
-        VersionReq::parse(">=0.2,<0.5.0-alpha").expect("valid version requirement"),
-        Version::parse("0.4.0").expect("valid version"),
-        migrate_0_2_to_0_3,
-    )];
+    let version_from = "0.4"; // this is the version that we'd like to migrate from
```
msk, line 50: I don't think having these variables is a good idea. Eventually `migration` will contain many entries. In that case, what would `version_from` means? Each element in `migration` should have a different migrate-from version, and defining a variable only for one of them looks strange.

src/backends/postgres/de.rs
```diff
@@ -445,7 +445,7 @@ where
     }
 }
 
-/// Deserializes an element of a PostgreSQL array.
+/// Deserializes an element of a `PostgreSQL` array.
```
msk, line 448: This is wrong. Add (with git) clippy.toml to the top-level directory with the following content instead.

```toml
doc-valid-idents = [ "PostgreSQL" ]
```

msk (changes requested):

src/migration.rs
sehkone: I followed Min Shao's instructions without adapting them properly to this particular situation. However, I have now made adjustments to align with your intended outcome.

src/backends/postgres/de.rs
sehkone: I changed it back to the original version, without the quotation.

src/migration.rs
sehkone: Now, I understand what it exactly means.

msk: Sorry, I meant alpha.4, the version it becomes with this PR.

msk: > Does this imply that it needs to be modified every time a new version emerges?

Only when the new version introduces a breaking change.

sehkone: So, this means the minimum required version, doesn't it?

sehkone: > Sorry, I meant alpha.4, the version it becomes with this PR.

Applied.

msk: > So, this means the minimum required version, doesn't it?

I'm afraid the name is misleading. I need to change it to something else. It is the versions that use the exactly same database format as the current version.

sehkone:
@minshao assisted me in modifying the migration section, but I'm uncertain whether it meets @msk's requirements for `DATABASE_VERSION_REQ`. Would both of you review the changes and confirm that they are appropriate?

src/migration.rs
```diff
@@ -40,32 +41,42 @@ pub fn migrate_data_dir<P: AsRef<Path>>(data_dir: P, backup_dir: P) -> Result<()
     }
 
     let mut version = data_ver;
-    if compatible.matches(&version) {
+    if version >= break_version {
```
msk, line 44: We cannot guarantee API compatibility with future versions. For example, 0.5.0-alpha.5 may introduce a change incompatible to 0.5.0-alpha.4. However, this condition will accept the database created by 0.5.0-alpha.5 and try to read it only to cause a potentially-disastrous error. Going back to an older version is common during debugging and thus we should prevent an old version from trying to read an incompatible future version's data.

msk (approved):

PR #31, opened 2023-04-27
Add event_source(s) to database, adjust queries accordingly

minshao:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "review-database"
-version = "0.6.0"
+version = "0.7.0-alpha"
```
msk, line 3: The first alpha version should be 0.7.0-alpha.1.

msk (changes requested):

msk:
Please read the following explanation on the SQL statement in `attempt_outlier_upsert` (Line 281). See the last paragraph for a potential problem:

This PL/pgSQL statement performs a dynamic SQL query using the `EXECUTE` command. Let's break it down:

- `'SELECT MIN(i), $2 FROM UNNEST($1) i, UNNEST($3) j WHERE i <> $4'`: This is the SQL query that will be executed. It selects the minimum value (`MIN(i)`) from the unnested array `$1` (`_event_ids`) and the value of `$2` from the unnested array `$3` (`_event_sources`). The `WHERE` clause filters out the values from `$1` that are equal to `$4`.

- `INTO _event_ids_update, _event_sources_update`: This is where the results of the `SELECT` statement will be stored. The minimum value will be stored in `_event_ids_update` and the `$2` value will be stored in `_event_sources_update`.

- `USING _event_ids, _event_sources, _event_ids, (SELECT MIN(i) FROM UNNEST(_event_ids) i)`: This is where the actual values for `$1`, `$2`, `$3`, and `$4` are provided for the SQL query. Here, `$1` and `$3` are both `_event_ids`, `$2` is `_event_sources`, and `$4` is the minimum value of `_event_ids`.

The purpose of this statement is to find the minimum value in the `_event_ids` array that is not equal to the current minimum value, and pair it with the corresponding value in `_event_sources`. These values are then stored in `_event_ids_update` and `_event_sources_update`, respectively.

However, please note that if the intention is to select the `$2` value from the same position as the minimum `i`, this statement doesn't guarantee that. It simply selects the minimum `i` and a `$2` value from somewhere in `_event_sources`. Without more information about the structure of these arrays, it's difficult to be certain about the exact behavior.

src/backends/postgres/migrations/2023-04-13-211209_add_source/up.sql
```diff
@@ -0,0 +1,306 @@
...
+  IF $10 IS NOT NULL THEN
+    _score := $10;
+  END IF;
+
+  IF _event_ids IS NOT NULL THEN
+    _event_ids := array_cat($3, _event_ids);
+    _event_sources := array_cat($4, _event_sources);
+    IF array_length(_event_ids, 1) > _max_event_id_num THEN
+      LOOP
+        EXECUTE 'SELECT MIN(i), $2 FROM UNNEST($1) i, UNNEST($3) j WHERE i <> $4' INTO _event_ids_update, _event_sources_update USING _event_ids, _event_sources, _event_ids, (SELECT MIN(i) FROM UNNEST(_event_ids) i);
```
minshao, line 97: @atyoshioka May I ask why in migration/2023-04-03-171325_remove_refinery/up.sql line 464 `UNNEST` is used? Originally I think `event_ids` is like `ARRAY[[4,5,6],[7,8,9]]`, but it seems to be `ARRAY[4, 5, 6, 7, 8, 9]`, so is `UNNEST` necessary in the original function? Did I miss anything?

atyoshioka: I'm not familiar with Postgres functions, so I guess there would be a better approach, but without `unnest`, that line doesn't work.

```
review001=# SELECT MIN(i) FROM ARRAY[1,2,3] i;
ERROR:  syntax error at or near "ARRAY"
LINE 1: SELECT MIN(i) FROM ARRAY[1,2,3] i;
                           ^
review001=# SELECT MIN(i) FROM unnest(ARRAY[1,2,3]) i;
 min 
-----
   1
(1 row)

```

minshao:
It shouldn't be `SELECT MIN(i)` instead it should be `SELECT i`.
And then I think $3 is redundant, already removed it~

src/backends/postgres/migrations/2023-04-13-211209_add_source/up.sql
minshao: I see, thanks~

atyoshioka:
The purpose of `down.sql` is to revert all changes made by `up.sql`. When `down.sql` is applied, all the tables, functions, indexes etc must be the same as they were before the `up.sql` was applied.

minshao:
> The purpose of down.sql is to revert all changes made by up.sql. When down.sql is applied, all the tables, functions, indexes etc must be the same as they were before the up.sql was applied.

Definitely agree~ I tried `redo` command, it seems ok. Did you notice something I forget to revert?

atyoshioka:
> Definitely agree~ I tried `redo` command, it seems ok. Did you notice something I forget to revert?

Well, for example, the functions are simply dropped. 

```
DROP TRIGGER IF EXISTS event_ids_update_trigger ON model;
DROP FUNCTION IF EXISTS attempt_cluster_upsert;
DROP FUNCTION IF EXISTS attempt_event_ids_update;
DROP FUNCTION IF EXISTS attempt_outlier_delete;
DROP FUNCTION IF EXISTS attempt_outlier_upsert;
```

If you run `redo`, I believe `up.sql` is applied again. What you need to run is probably `diesel migration revert`.

minshao:
> > Definitely agree~ I tried `redo` command, it seems ok. Did you notice something I forget to revert?
> 
> Well, for example, the functions are simply dropped.
> 
> ```
> DROP TRIGGER IF EXISTS event_ids_update_trigger ON model;
> DROP FUNCTION IF EXISTS attempt_cluster_upsert;
> DROP FUNCTION IF EXISTS attempt_event_ids_update;
> DROP FUNCTION IF EXISTS attempt_outlier_delete;
> DROP FUNCTION IF EXISTS attempt_outlier_upsert;
> ```
> 
> If you run `redo`, I believe `up.sql` is applied again. What you need to run is probably `diesel migration revert`.

According to [diesel](https://diesel.rs/guides/getting-started) "It’s a good idea to make sure that down.sql is correct. You can quickly confirm that your down.sql rolls back your migration correctly by redoing the migration", therefore, I thought as long as I run `redo` and it is fine that means both application and reversion of the migration are fine.

src/backends/postgres/migrations/2023-04-13-211209_add_source/up.sql
```diff
@@ -0,0 +1,306 @@
...
+  IF $10 IS NOT NULL THEN
+    _score := $10;
+  END IF;
+
+  IF _event_ids IS NOT NULL THEN
+    _event_ids := array_cat($3, _event_ids);
+    _event_sources := array_cat($4, _event_sources);
+    IF array_length(_event_ids, 1) > _max_event_id_num THEN
+      LOOP
+        EXECUTE 'SELECT i,  j FROM $1  i, $2 j WHERE i <> $3' INTO _event_ids_update, _event_sources_update USING _event_ids, _event_sources, (SELECT MIN(i) FROM _event_ids i);
```
atyoshioka, line 97: I believe you forgot `unnest`:

```
(SELECT MIN(i) FROM _event_ids i)
```

Also, is this a valid `SELECT` statement? Did you confirm that this code works? Because I got an error in my environment saying "syntax error at or near "$1""

atyoshioka: I meant this `SELECT` statement:

```
'SELECT i,  j FROM $1  i, $2 j WHERE i <> $3'
```

minshao: Thanks for catching it!

I fixed it and simplified the statement, and below is a mini test to show the code and the result:

``` SQL
review=# CREATE OR REPLACE FUNCTION test()
RETURNS TABLE(id INTEGER, name TEXT) AS $$
DECLARE
   a INTEGER[] := ARRAY[1, 2, 1];
   b TEXT[] := ARRAY['apple', 'banana', 'orange'];
BEGIN
   RETURN QUERY SELECT a[i], b[i] 
   FROM generate_series(1, array_length(a, 1)) i 
   WHERE a[i] <> (SELECT MIN(j) FROM unnest(a) j);
END;
$$ LANGUAGE plpgsql;
CREATE FUNCTION
review=# select test();
    test    
------------
 (2,banana)
(1 row)

review=# 
```

minshao:
> Yeah, that's just a one way to test that `down.sql` works correctly. But `down.sql` is to undo everything in `up.sql`. That how we used migrations before we switched to tokio_postgres.

I'm guessing you are referring to the functions that I dropped but didn't create: `attempt_cluster_upsert` etc. I've fixed it by copying the function to `down.sql`. Hope that's the way to do it, let me know if I need to fix anything. A pointer/comment right at the code would be so appreciated~ Thanks!

src/backends/postgres/migrations/2023-04-13-211209_add_source/up.sql
atyoshioka: You seem to forget to update line 97. As for the code you updated, `SELECT _event_ids[i], _event_sources[i]` returns `BIGINT` and `TEXT` but the data type of `_event_ids_update ` and `_event_sources_update ` are `BIGINT[]` and `TEXT[]`. Type mismatches. I believe if you update your test function to use `INTO` after the `WHERE` clause, it doesn't work.

```
      SELECT _event_ids[i], _event_sources[i] 
      FROM generate_series(1, array_length(_event_ids, 1)) i 
      WHERE _event_ids[i] <> (SELECT MIN(j) FROM unnest(_event_ids) j)
      INTO _event_ids_update, _event_sources_update;
```

atyoshioka:
> I'm guessing you are referring to the functions that I dropped but didn't create: `attempt_cluster_upsert` etc. I've fixed it by copying the function to `down.sql`. Hope that's the way to do it, let me know if I need to fix anything. A pointer/comment right at the code would be so appreciated~ Thanks!

Thanks for the update!

src/backends/postgres/migrations/2023-04-13-211209_add_source/up.sql
minshao:
``` SQL
review=# CREATE OR REPLACE FUNCTION test()
RETURNS INTEGER[] AS $$ DECLARE
   a INTEGER[] := ARRAY[1, 2, 3, 1];
   b TEXT[] := ARRAY['apple', 'banana', 'mango', 'orange'];
   c INTEGER[];
   d TEXT[];
BEGIN
   SELECT ARRAY_AGG(ac), ARRAY_AGG(bc) INTO c, d FROM (
      SELECT a[i] as ac, b[i] as bc FROM generate_series(1, array_length(a, 1)) i
      WHERE a[i] <> (SELECT MIN(j) FROM unnest(a) j)
   ) t;
   RETURN c;
END;
$$ LANGUAGE plpgsql;
CREATE FUNCTION
review=# select test();
 test  
-------
 {2,3}
(1 row)
```

Above is the updated SQL statement test

atyoshioka: If you already confirm that the code works as expected, then it is ready to be merged. (I have no idea what the above code is trying to achieve. To me, it simply removes the minimus value from `a` but `b` is also manipulated for some reason. I don't know the relationship between `event_ids` and `event_sources`)

minshao: The intention is to remove the minimum element in `a` and the element in `b` that's at the same position. Since (`event_ids[i]`, `event_sources[i]`) defines an event, it has to be deleted/added at the same time. And in this case, we want to remove the earliest event (`MIN(event_ids)`). So we need to remove (`event_ids[i]`, `event_sources[i]`) when `event_ids[i] == MIN(event_ids)`. Hope that explains it~

msk (approved):

Issue #56, opened 2023-05-16
Extend `HttpThreatFields` Struct with Additional Fields 


syncpark:
#### Issue Description:
refer petabi/review-database#49

The current HttpThreatFields has basic set of fields:

```rust
pub(super) struct HttpThreatFields {
    pub event_id: u64,
    pub time: DateTime<Utc>,
    pub source: String,
    pub src_addr: IpAddr,
    pub src_port: u16,
    pub dst_addr: IpAddr,
    pub dst_port: u16,
    pub proto: u8,
    pub duration: i64,
    pub host: String,
    pub content: String,
    pub db_name: String,
    pub rule_id: u32,
    pub cluster_id: usize,
    pub attack_kind: String,
    pub confidence: f32,
}
```
Extend HttpThreatFields like follows

```rust
pub(super) struct HttpThreatFields {
    pub event_id: u64,
    pub time: DateTime<Utc>,
    pub source: String,
    pub src_addr: IpAddr,
    pub src_port: u16,
    pub dst_addr: IpAddr,
    pub dst_port: u16,
    pub proto: u8,
    pub duration: i64,
    pub method: String,
    pub host: String,
    pub uri: String,
    pub referer: String,
    pub version: String,
    pub user_agent: String,
    pub request_len: usize,
    pub response_len: usize,
    pub status_code: u16,
    pub status_msg: String,
    pub username: String,
    pub password: String,
    pub cookie: String,
    pub content_encoding: String,
    pub content_type: String,
    pub cache_control: String,
    pub db_name: String,
    pub rule_id: u32,
    pub cluster_id: usize,
    pub attack_kind: String,
    pub confidence: f32,
}
```

#### Tasks
* Extend the HttpThreatFields struct with the new fields.
* Update the function(s) that instantiate and use HttpThreatFields to accommodate the new fields.
* Add a database migration function to handle the new fields in existing data, filling the new fields with default values.
* Create unit tests to verify the correct functionality of the migration

PR #63, opened 2023-05-17
Extend HttpThreat fields and add matched_to field

syncpark:
petabi/review-database#56

* Extend `HttpThreatFields` to use all of it's original fields
* Add `matched_to` field to contain the matched patterns


src/migration.rs
```diff
@@ -578,6 +583,106 @@ fn migrate_0_7_to_0_9<P: AsRef<Path>>(path: P, backup: P) -> Result<()> {
     Ok(())
 }
 
+/// Migrates the database from 0.9 to 0.10.
```
msk, line 586: Do you mean to 0.11? We've already released 0.10.1.

sehkone:
Please ensure that this feature is incorporated into tomorrow's release. We need to provide this tomorrow to the client.

PR #78, opened 2023-05-26
Fix migration bug for HttpEventFields

syncpark:
Fix migration bug for HttpEventFields data


syncpark:
> Does #79 fix this issue?

Yes #79 works fine. I'll drop this PR. Thanks.

PR #82, opened 2023-05-31
Add `Store::pretrained_model` to get model from corresponding file

### LINKED ISSUE #74: Add `Store::pretrained_model` Function to review-database Crate
## Summary:
We propose adding a new public function, `Store::pretrained_model`, to the review-database crate. This function will receive a model name as an argument and return a `PretrainedModel` struct.

## Description:

The purpose of the `Store::pretrained_model` function is to retrieve a pre-trained model from the store based on its name. The function should take a model name as an argument and return a `PretrainedModel` struct, defined as `pub struct PretrainedModel(pub Vec<u8>)`. The content of the struct (`Vec<u8>`) should be the content of a file named using the given model name followed by a hyphen and a UNIX timestamp (seconds since the epoch), such as "ransomware-dns-1684954208". 

These candidate model files should be searched in the "pretrained" directory located under the `path` that was given to `Store::new` when creating a `Store` instance.

If there are multiple files that start with the given model name, the function should load the one with the latest timestamp. Note that the model name may contain a hyphen.

## Proposed Changes:

1. Add the `Store::pretrained_model` function to the review-database crate.
2. Implement the function to receive a model name and return a `PretrainedModel` struct.
3. Search for model files in the "pretrained" directory under the `path` given during the `Store` instance creation.
4. In cases where multiple files start with the model name, ensure the function loads the one with the latest timestamp.

## Acceptance Criteria:

1. The `Store::pretrained_model` function is successfully added to the review-database crate and functions as expected.
2. The function correctly takes a model name and returns a `PretrainedModel` struct.
3. The function accurately retrieves and loads the model file with the latest timestamp in cases where there are multiple files with the same starting name from the correct directory.

minshao:
Resolves #74

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,13 +62,14 @@ pub use self::types::{
     HostNetworkGroup, ModelIndicator, PacketAttr, Response, ResponseKind, Ti, TiCmpKind,
     TriagePolicy, ValueKind,
 };
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use backends::Value;
 use bb8_postgres::{
     bb8,
     tokio_postgres::{self, types::Type},
 };
 pub use rocksdb::backup::BackupEngineInfo;
+use std::io;
 use std::path::{Path, PathBuf};
 use thiserror::Error;
 
@@ -97,9 +98,11 @@ const EXCLUSIVE: bool = true;
 pub struct Store {
     states: StateDb,
     backup: PathBuf,
+    pretrained: PathBuf,
 }
 
 impl Store {
+    const DEFAULT_PRETRAINED: &str = "pretrained";
     /// Opens a new key-value store and its backup.
     ///
     /// # Errors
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -329,6 +354,48 @@ impl Store {
...
+            if let Some((cur_ts, _)) = &most_recent {
+                if ts > *cur_ts {
+                    most_recent = Some((ts, entry));
+                }
+            } else {
+                most_recent = Some((ts, entry));
+            }
+        }
+    }
+    most_recent.ok_or(anyhow!("Fail to locate {name:?} under {dir:?}"))
```
msk, line 396: Using [`Path::display`](https://doc.rust-lang.org/std/path/struct.Path.html#method.display) is more conventional way to print a path.

msk (changes requested):

msk (approved):

PR #88, opened 2023-06-05
add migration log message

syncpark:
Add log message for migration.
In the case of a large DB, the migration process takes too much time, and users may misunderstand that it is malfunctioning.


msk (changes requested):
Please update CHANGELOG for any user-visible change.

msk (approved):

PR #94, opened 2023-06-14
Backup

### LINKED ISSUE #87: `backup::restore` is not working
Currently, the restore function of `review-database` does not seem to work properly.
For the restore test, after running review, I proceeded as follows via `graphql`.

1. Call `insertBlockNetwork` 5 times to add data to rocksdb.
2. Call `blockNetworkList `to verify that 5 pieces of data have been added.
3. Call `backup` to perform a backup of rocksdb, and verify that the backup data is added to the specified backup path.
4. Call `removeBlockNetworks `to delete all blockNetworks.
5. Call `blockNetworkList` to verify that all data has been deleted.
6. Call `restoreFromLatestBackup`, and then call `blockNetworkList` to verify the change in data.
7.  Re-run review and then call `blockNetworkList` to verify the change in data.

In my tests, the 5-block network was not restored.

minshao:
Resolves #87

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,59 @@ file is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
 this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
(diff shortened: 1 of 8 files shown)


msk (changes requested):
Please update CHANGELOG.

msk (approved):

Issue #71, opened 2023-05-23
Enhance `backup::create` to support PostgreSQL backup


msk:
**Description:**

Currently, the `backup::create` function creates a backup for RocksDB databases. To increase our application's versatility and usefulness, we want to extend this function to also support backups for PostgreSQL databases. 

This enhancement requires the function to run the `pg_dump` external command to create a PostgreSQL backup, storing the resulting file in the backup directory. Additionally, the filename must contain the same backup ID as the RocksDB backup created within the same function call. This is to ensure that the matching RocksDB/PostgreSQL backup pair can be easily selected when a user decides to restore from backups.

**Acceptance Criteria:**

1. `backup::create` should be able to create a backup for PostgreSQL databases.
2. The function should use the `pg_dump` command to create the PostgreSQL backup.
3. The backup file should be stored in the backup directory.
4. The filename of the PostgreSQL backup should contain the same backup ID as the RocksDB backup created within the same function call to `backup::create`.

msk:
Obsoleted by #434.

Issue #80, opened 2023-05-26
Enhance `backup::restore` function to support PostgreSQL in addition to RocksDB


msk:
## Description

Currently, the `backup::restore` function is designed to restore backups for RocksDB only. This issue aims to enhance this function to support PostgreSQL as well. 

This change would involve not only restoring the state of the RocksDB store but also the PostgreSQL database from its corresponding backup. It's proposed to use `psql` command for restoring the PostgreSQL backup.

## Tasks

1. Modify the `backup::restore` function to restore both RocksDB and PostgreSQL databases.
2. Ensure the function identifies the correct backup files for each database type.
3. Use the `psql` command to restore the PostgreSQL backup.
4. Thoroughly test the new functionality, ensuring that it can correctly restore the state of both the RocksDB store and the PostgreSQL database.

## Acceptance Criteria

1. The `backup::restore` function is able to restore both RocksDB and PostgreSQL backups.
2. All tests pass without causing any regression.

kimhanbeom:
@msk @syncpark

I am currently working on an implementation, referring to the restoration part of https://github.com/aicers/cli. 
In order to use this feature, I need to call `sudo_run_command` and that requires root privileges.

To solve this, it looks like we need to choose one of 2 ways.

1. Run review as root to perform the command.
2. Add `sudo_run_command` to `roxy` to perform the command via roxy.

msk:
Which step requires the root privilege?

kimhanbeom:
The part of the `postgres_restore_docker` function that moves data from the existing postgresql to tmp requires permissions.

msk:
Could we consider the option of creating a temporary directory that's writable for the REview process instead? This way, we can circumvent the need for superuser permissions during the process of moving data from existing postgresql to a temporary directory.

It's generally a bad practice to run applications with root privileges unless absolutely necessary, due to the significant security risks involved. When a process runs with root privileges, it has unrestricted access to the system. This means it can read, write, and delete every file on your system, and it can execute any command. If any vulnerabilities exist within the application, an attacker could exploit them to gain full control over the system.

msk:
Obsoleted by #434.

Issue #76, opened 2023-05-26
Enhance `backup::list` Function to Include PostgreSQL Backups


msk:
## Summary:
Building on the work done in Issue #73, this issue focuses on enhancing the `backup::list` function to include PostgreSQL backups. Each RocksDB backup is paired with a PostgreSQL backup, and the `size` field of each `BackupInfo` should reflect the combined size of both backups. This task depends on the completion of Issue #71.

## Description:

The `backup::list` function implemented in the `review-database` crate currently retrieves information about RocksDB backups as described in Issue #73. This function should be enhanced to include PostgreSQL backups, which correspond to each RocksDB backup.

The returned `Vec<BackupInfo>` should include the total size of both the RocksDB and PostgreSQL backups in the `size` field of each `BackupInfo`.

Note that work on this issue should start only after the completion of Issue #71, as the implementation relies on the changes defined there.

## Proposed Changes:

1. Enhance the `backup::list` function to include information about PostgreSQL backups.
2. Update the `size` field of each `BackupInfo` to reflect the combined size of the corresponding RocksDB and PostgreSQL backups.

## Acceptance Criteria:

1. The `backup::list` function returns a `Vec<BackupInfo>` that includes information about both RocksDB and PostgreSQL backups.
2. The `size` field of each `BackupInfo` accurately reflects the combined size of the corresponding RocksDB and PostgreSQL backups.

msk:
Obsoleted by #434.

PR #105, opened 2023-06-23
[WIP] Enhance backup logic

kimhanbeom:
- Added new module 'postgres' to support postgresql.
- Enhance `backup::list`/`backup::create`/`backup::restore`/`backup::recovery` function to support postgresql.
- Added `backup::purge_old_backups` for apply immediately after `num_backups_to_keep` is changed.

Closes: #71, #76, #80

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@ file is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
 this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Introduced a new module `postgres`. This module was added to allow the
+  existing backup `create`/`list`/`restore`/`recover` function to also support
+  postgresql DBs.
+- Added `backup::purge_old_backups` for apply immediately after
+  `num_backups_to_keep` is changed.
+
+### Changed
+
+- Enhanced to also support postgresql for backup `create`/`list`/`restore`
+  /`recover`.
+
 ## [0.15.2] - 2023-07-06
 
 ### Added
@@ -422,6 +437,7 @@ leading to a more streamlined system.
 
 - An initial version.
 
+[Unreleased]: https://github.com/petabi/review-database/compare/0.15.2...main
 [0.15.2]: https://github.com/petabi/review-database/compare/0.15.1...0.15.2
 [0.15.1]: https://github.com/petabi/review-database/compare/0.15.0...0.15.1
 [0.15.0]: https://github.com/petabi/review-database/compare/0.14.1...0.15.0
--- a/src/backup.rs
+++ b/src/backup.rs
@@ -1,13 +1,21 @@
 //! Database backup utilities.
+mod postgresql;
 
-use crate::Store;
-use anyhow::Result;
+#[allow(clippy::module_name_repetitions)]
+pub use self::postgresql::BackupConfig;
+use self::postgresql::{create_backup_path, purge_old_postgres_backups};
+use crate::{
+    backup::postgresql::{postgres_backup, postgres_backup_list, postgres_restore},
+    Store,
+};
+use anyhow::{anyhow, Result};
 use chrono::{DateTime, TimeZone, Utc};
 use rocksdb::backup::BackupEngineInfo;
 use std::{sync::Arc, time::Duration};
 use tokio::sync::{Notify, RwLock};
-use tracing::{info, warn};
+use tracing::info;
 
+#[derive(Debug, Clone)]
 #[allow(clippy::module_name_repetitions)]
 pub struct BackupInfo {
     pub id: u32,
@@ -19,7 +27,10 @@ impl From<BackupEngineInfo> for BackupInfo {
     fn from(backup: BackupEngineInfo) -> Self {
         Self {
             id: backup.backup_id,
-            timestamp: Utc.timestamp_nanos(backup.timestamp),
+            timestamp: Utc
+                .timestamp_opt(backup.timestamp, 0)
+                .single()
+                .expect("Invalid timestamp value"),
             size: backup.size,
         }
     }
@@ -29,8 +40,8 @@ impl From<BackupEngineInfo> for BackupInfo {
 #[allow(clippy::module_name_repetitions)]
 pub async fn schedule_periodic(
     store: Arc<RwLock<Store>>,
+    backup_cfg: Arc<RwLock<BackupConfig>>,
     schedule: (Duration, Duration),
-    backups_to_keep: u32,
     stop: Arc<Notify>,
 ) {
     use tokio::time::{sleep, Instant};
@@ -43,11 +54,10 @@ pub async fn schedule_periodic(
         tokio::select! {
             () = &mut sleep => {
                 sleep.as_mut().reset(Instant::now() + duration);
-                let _res = create(&store, false, backups_to_keep);
+                let _res = create(&store, false, &backup_cfg);
             }
             _ = stop.notified() => {
-                info!("creating a database backup before shutdown");
-                let _res = create(&store, false, backups_to_keep);
+                let _res = create(&store, false, &backup_cfg);
                 stop.notify_one();
                 return;
             }
@@ -61,23 +71,28 @@ pub async fn schedule_periodic(
 /// # Errors
 ///
 /// Returns an error if backup fails.
-pub async fn create(store: &Arc<RwLock<Store>>, flush: bool, backups_to_keep: u32) -> Result<()> {
-    // TODO: This function should be expanded to support PostgreSQL backups as well.
+pub async fn create(
+    store: &Arc<RwLock<Store>>,
+    flush: bool,
+    backup_cfg: &Arc<RwLock<BackupConfig>>,
+) -> Result<()> {
     info!("backing up database...");
-    let res = {
-        let mut store = store.write().await;
-        store.backup(flush, backups_to_keep)
-    };
-    match res {
-        Ok(_) => {
-            info!("backing up database completed");
-            Ok(())
-        }
-        Err(e) => {
-            warn!("database backup failed: {:?}", e);
-            Err(e)
+    {
+        let num_of_backups = { backup_cfg.read().await.num_of_backups };
+        let mut backup_store = store.write().await;
+        if let Err(e) = backup_store.backup(flush, num_of_backups) {
+            return Err(anyhow!("failed to create key-value database backup: {e:?}"));
         }
     }
+
+    let backup_id_list = backup_id_list(store).await?;
+    let backup_cfg = backup_cfg.read().await;
+    if let Err(e) = postgres_backup(&backup_cfg, backup_id_list) {
+        return Err(anyhow!(
+            "failed to create relational database backup: {e:?}"
+        ));
+    }
+    Ok(())
 }
 
 /// Lists the backup information of the database.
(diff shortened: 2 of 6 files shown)


src/backup/postgresql.rs
```diff
@@ -0,0 +1,441 @@
...
+    Ok(())
+}
+
+/// # Errors
+/// * restore command not found
+/// * fail to restore
+fn postgres_restore_docker(cfg: &BackupConfig, dump: &str) -> Result<()> {
+    // copy current database folder (For manual restore)
+    if let Err(e) = run_command("cp", None, &["-r", &cfg.postgres_db_path, DEFAULT_TEMP_DIR]) {
+        return Err(anyhow!("fail to remove old database. {:?}", e));
```
syncpark, line 275: fail to remove old database -> fail to `copy` database
copy or backup

src/backup/postgresql.rs
```diff
@@ -0,0 +1,441 @@
+use super::BackupInfo;
+use anyhow::{anyhow, Context, Result};
+use flate2::{read::GzDecoder, write::GzEncoder, Compression};
+use regex::Regex;
+use std::collections::HashSet;
+use std::fs::{self, File};
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+const DEFAULT_PATH_ENV: &str = "/usr/sbin:/usr/bin:/sbin:/bin";
```
syncpark, line 10: DEFAULT_PATH_ENV는 docker, cp 등의 커맨드를 찾기 위한 기본 경로입니다.
만일 이들이 `/usr/bin` 아래에 모두 모여있다면 그리고 그게 현재 제품을 설치한 모든 고객서버에서 똑같은 상황이라면, 다른 경로는 필요가 없습니다.

현재 코드에서 사용하는 docker, cp는 `/usr/bin` 혹은 `/bin`에 설치됩니다. 나머지는 삭제하는게 좋겠습니다.
다만 만일의 경우를 대비하여 확실하게 에러 메시지를 저장하거나 커맨드가 해당 경로에 존재하는지 확인할 필요는 있겠습니다.

syncpark (changes requested):

kimhanbeom:
해당 사항 수정 완료 했습니다.

syncpark (changes requested):

CHANGELOG.md
```diff
@@ -5,6 +5,21 @@ file is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
...
+
+### Changed
+
+- Enhanced to also support postgresql for backup `create`/`list`/`restore`
+  /`recover`.
+
+### Added
+
+- Introduced a new module `postgres`. This module was added to allow the
+  existing backup `creat`/`list`/`restore`/`recover` function to also support
```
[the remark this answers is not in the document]
kimhanbeom: 수정 완료 했습니다.

syncpark (approved):

src/backup/postgresql.rs
```diff
@@ -0,0 +1,471 @@
...
+}
+
+/// # Errors
+/// * dump command not found
+/// * fail to dump
+fn postgres_dump_docker(cfg: &BackupConfig, to: &str) -> Result<()> {
+    let dump_file = format!("{DEFAULT_POSTGRES_TEMP_DIR}/{DEFAULT_POSTGRES_DUMP_FILE}");
+    let args = vec!["exec", "-i", &cfg.container, "/bin/rm", "-f", &dump_file];
+
+    run_command("docker", None, &args)
```
msk, line 249: The current implementation assumes the presence of Docker, which may not be universally available or preferred in all deployment scenarios. For maximum flexibility, the `review-database` should be able to support various methods of running `pg_dump` and shouldn't rely on Docker or any specific containerization technology.

Therefore, we recommend making this more flexible by allowing the user to provide the full path to the `pg_dump` command at runtime. This could be achieved through a configuration setting or a command-line argument, for example. By using the full path, `review-database` doesn't need to manage or modify the `PATH` environment variable itself, which makes it more resilient to differences in system configuration and more reliable in terms of finding the correct `pg_dump` command.

The command could be something like `/usr/bin/pg_dump` or `/usr/local/bin/docker exec -i [container] pg_dump`, depending on the user's specific setup.

This would allow for greater flexibility and accommodate a wider range of deployment scenarios.

src/backup/postgresql.rs
```diff
@@ -0,0 +1,471 @@
...
+    remove_tmpdir_all(&tmpdir)?;
+
+    Ok(data_backup_path)
+}
+
+/// # Errors
+/// * dump command not found
+/// * fail to dump
+fn postgres_dump_docker(cfg: &BackupConfig, to: &str) -> Result<()> {
+    let dump_file = format!("{DEFAULT_POSTGRES_TEMP_DIR}/{DEFAULT_POSTGRES_DUMP_FILE}");
```
msk, line 246: `DEFAULT_POSTGRES_TEMP_DIR` is being used to construct the path for the database dump file. While this works, it may not be the best approach because it assumes a particular directory structure and relies on hardcoded paths.

Instead, I recommend using the `tempfile` crate, which is a library designed for dealing with temporary files in Rust. This crate creates files and directories in the location returned by `std::env::temp_dir()`, which respects the operating system's conventions and is generally a good place to store temporary files.

src/backup.rs
```diff
@@ -61,104 +73,164 @@ pub async fn schedule_periodic(
...
+    };
+    let mut backup_list: Vec<BackupInfo> = backup_list
+        .into_iter()
+        .map(std::convert::Into::into)
+        .collect();
+    if let Err(e) = postgres_backup_list(backup_path, &mut backup_list) {
+        return Err(anyhow!(
+            "failed to add list information from a relational database: {e:?}"
+        ));
     }
```
minshao, line 119:
```Rust
     let res = {
         let store = store.read().await;
         store.get_backup_info()
     };
     let mut backup_list = match res {
         Ok(backup_list) => {
             info!("generate database backup list");
             backup_list
                 .into_iter()
                 .map(std::convert::Into::into)
                 .collect()
         }
         Err(e) => {
             warn!("failed to generate backup list: {:?}", e);
             return Err(e);
         }
    }
    if let Err(e) = postgres_backup_list(backup_path, &mut backup_list) {
         return Err(anyhow!(
             "failed to add list information from a relational database: {e:?}"
         ));
    }

    

```
Read lock should be released as soon as possible. i.e. Add the `postgres_backup_list` call after the lock has been released

src/backup/postgresql.rs
```diff
@@ -0,0 +1,471 @@
...
+        return Err(anyhow!("No database found"));
+    }
+
+    //create backup folder
+    let Ok(data_backup_path) = create_backup_path(cfg) else {
+        return Err(anyhow!("Backup folder creation failed"));
+    };
+
+    // mkdir default temporary folder
+    if !Path::new(DEFAULT_HOST_TEMP_DIR).exists() && fs::create_dir(DEFAULT_HOST_TEMP_DIR).is_err()
```
msk, line 209: As `review-database` should not make assumptions about the environment where `REview` is running, this default directory path may not always be applicable or suitable. It could lead to issues on different operating systems or environments where the path is not accessible or does not exist.

Therefore, I suggest altering the implementation to receive this temporary directory path as an argument from `REview`. This would make the feature more flexible and adaptable to different environments, as `REview` would have more knowledge of the suitable paths in its running environment.

msk (changes requested):

src/backup.rs
```diff
@@ -61,104 +73,164 @@ pub async fn schedule_periodic(
...
-            Ok(())
-        }
-        Err(e) => {
-            warn!(
-                "failed to restore database from backup {:?}: {:?}",
-                backup_id, e
-            );
-            Err(e)
+    let mut store = store.write().await;
+    if let Err(e) = store.restore_from_backup(backup_id) {
```
minshao, line 150: please drop `store` as soon as restore is completed

src/backup.rs
```diff
@@ -61,104 +73,164 @@ pub async fn schedule_periodic(
...
-        store.recover()
-    };
 
-    match res {
-        Ok(_) => {
-            info!("database recovered from backup");
-            Ok(())
-        }
+    let mut store = store.write().await;
+    let recovery_id = match store.recover() {
```
minshao, line 206: please drop `store` as soon as `recover` is called

src/backup/postgresql.rs
kimhanbeom: `DEFAULT_POSTGRES_TEMP_DIR` means the tmp folder of the `container` in `Postgres`. Save the result of `pg_dump` to the tmp of the container, copy the saved dump file to the path provided by the host, and finally backup the copied file.

src/backup.rs
```diff
@@ -19,7 +29,7 @@ impl From<BackupEngineInfo> for BackupInfo {
     fn from(backup: BackupEngineInfo) -> Self {
         Self {
             id: backup.backup_id,
-            timestamp: Utc.timestamp_nanos(backup.timestamp),
+            timestamp: Utc.timestamp_nanos(backup.timestamp * BACKUP_TIMESTAMP_CONV_UNIT),
```
minshao, line 32: using `Utc::from_timestamp_opt(backup.timestamp, 0).expect("some message")` would be a better choice IMHO.

src/backup.rs
```diff
@@ -43,11 +53,13 @@ pub async fn schedule_periodic(
         tokio::select! {
             () = &mut sleep => {
                 sleep.as_mut().reset(Instant::now() + duration);
-                let _res = create(&store, false, backups_to_keep);
+                let backup_cfg= backup_cfg.read().await.clone();
+                let _res = create(&store, false, &backup_cfg);
```
minshao, line 57: The lock of `backup_cfg` will be hold while trying to get `store` lock. It's very likely to cause a deadlock.
A safer option is to pass `backup_cfg` to `create` function which will have 2 phases: 1. get the lock for `store` and backup rocksdb, release the lock; 2. get the lock for `backup_cfg` and backup for `postgres` and release the lock;

src/backup.rs
```diff
@@ -61,104 +73,175 @@ pub async fn schedule_periodic(
 /// # Errors
 ///
 /// Returns an error if backup fails.
-pub async fn create(store: &Arc<RwLock<Store>>, flush: bool, backups_to_keep: u32) -> Result<()> {
-    // TODO: This function should be expanded to support PostgreSQL backups as well.
+pub async fn create(
+    store: &Arc<RwLock<Store>>,
+    flush: bool,
+    backup_cfg: &BackupConfig,
```
minshao, line 79: As mentioned earlier, switching to `&Arc<RwLock<BackupConfig>>`

src/backup.rs
```diff
@@ -61,104 +73,175 @@ pub async fn schedule_periodic(
...
 ///
 /// # Errors
 ///
 /// Returns an error if the restore operation fails.
-pub async fn restore(store: &Arc<RwLock<Store>>, backup_id: Option<u32>) -> Result<()> {
-    // TODO: This function should be expanded to support PostgreSQL backups as well.
-    info!("restoring database from {:?}", backup_id);
+pub async fn restore(
+    store: &Arc<RwLock<Store>>,
+    cfg: &BackupConfig,
```
minshao, line 136: Similar to `create` function: use `cfg: &Arc<RwLock<BackupConfig>>` and separate the 2 restoring phases

src/backup.rs
```diff
@@ -61,104 +73,175 @@ pub async fn schedule_periodic(
...
+}
+
+/// Remove older backups based on the number of backups retained.
+///
+/// # Errors
+///
+/// Returns an error if removing old backup fails.
+pub async fn purge_old_backups(
+    store: &Arc<RwLock<Store>>,
+    backup_cfg: &BackupConfig,
```
minshao, line 188: Similar to create function: use cfg: &Arc<RwLock<BackupConfig>> and separate the 2 restoring phases

src/backup.rs
```diff
@@ -61,104 +73,175 @@ pub async fn schedule_periodic(
...
 }
 
 /// Restores the database from a backup with the specified ID.
 ///
 /// # Errors
 ///
 /// Returns an error if the restore operation fails.
-pub async fn recover(store: &Arc<RwLock<Store>>) -> Result<()> {
-    // TODO: This function should be expanded to support PostgreSQL backups as well.
+pub async fn recover(store: &Arc<RwLock<Store>>, cfg: &BackupConfig) -> Result<u32> {
```
minshao, line 210: Similar to create function: use cfg: &Arc<RwLock<BackupConfig>> and separate the 2 restoring phases

src/backup.rs
```diff
@@ -61,104 +73,175 @@ pub async fn schedule_periodic(
...
 }
 
 /// Lists the backup information of the database.
 ///
 /// # Errors
 ///
 /// Returns an error if backup list fails to create
-pub async fn list(store: &Arc<RwLock<Store>>) -> Result<Vec<BackupInfo>> {
-    // TODO: This function should be expanded to support PostgreSQL backups as well.
+pub async fn list(store: &Arc<RwLock<Store>>, backup_path: &Path) -> Result<Vec<BackupInfo>> {
```
minshao, line 103: although I haven't seen the usage of the function yet, it's better to keep it inline with other places and pass the `&Arc<RwLock<BackupConfig>>` for a unified interface.

src/backup.rs
kimhanbeom: For `create` and `purge_old_backups`, rocksdb require `num_of_backups`, so eventually the lock on the `BackupConfig` will take until the end of the function.

So, what if I modified it like this?

```
   let backup_cfg = { 
            backup_cfg.read().await.clone() 
    };
```

Previously, passing a reference to a value using clone would still hold the lock, but this would release the lock immediately.

minshao: In short, I don't think that would do the trick. 

The reason for using lock is to guarantee that at, any given time, if someone is writing, no one else can read or write. By doing what you explained, a read while write might happen for the file listed under the directory of `backup_cfg`.

minshao: Let me know if there's anything I can help with~

kimhanbeom:
@minshao 
We are currently working on a simultaneous release to all repositories. We had a meeting with sehkone and decided not to include the improved backup in this release, So I'll change it to WIP until the next version is released.

syncpark (approved):

syncpark:
@kimhanbeom rebase 해주세요

msk:
Closed per discussion in #206.

PR #110, opened 2023-07-03
Add `CryptocurrencyMiningPool` event & Modify event fields

BLYKIM:
- The `src_port` is not provided in the `FtpBruteForce`, `LdapBruteForce` events.
- Modify `LdapPlainText` fields to the appropriate event.
- Added `is_internal` field to `FtpBruteForce`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Added a `CryptocurrencyMiningPool` event with `dns` protocol.
+
 ### Changed
 
 - Renamed `ExternalDDos` to `ExternalDdos` and `ExternalDDosFields` to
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -20,6 +24,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   code if you have been using the old naming convention. We apologise for any
   inconvenience this may cause, but we believe this change will bring greater
   consistency and readability to the codebase.
+- Removed `src_port` field from `FtpBruteForce` and `LdapBruteForce` events.
+- Modify `LdapPlainText` fields.
```
msk, line 28: This CHANGELOG entry could benefit from more detail. Given that the review-database crate is used by other projects, it's important to ensure that our CHANGELOG entries provide sufficient information for developers of those projects to understand what has been changed, why it has been changed, and how it might affect their usage of our crate.

For example, if there are breaking changes, we need to clearly state those so that dependent projects can plan their upgrades accordingly. Or if we've made performance improvements, developers may want to upgrade even if they don't need to for compatibility reasons.

A more informative CHANGELOG entry might look something like this:

- Modified the `LdapPlainText` fields to improve [functionality/security/performance/etc.]. Changes include [description of specific changes]. This change [will/will not] require updates in dependent projects due to [reason].

Would you be able to revise your CHANGELOG entry accordingly?

msk (changes requested):

src/event.rs
```diff
@@ -116,6 +117,9 @@ pub enum Event {
 
     /// Plain text password is used for the LDAP connection.
     LdapPlainText(LdapPlainText),
+
+    /// An event that occurs when it is determined that there is a connection to a cryptocurrency mining network
+    Cryptocurrency(Cryptocurrency),
```
msk, line 122: While I understand the rationale behind the choice of this term, considering that the event is specifically triggered by identifying traffic to crypto mining pool sites, it might be more descriptive to name the event type "CryptocurrencyMiningPool".

This name suggests a more precise context for the event and provides greater clarity for those reviewing the database events in the future. By specifying that this event is associated with "mining pool" activities rather than a broader "cryptocurrency" context, we can better segment and analyze our data.

msk (changes requested):

syncpark (approved):

syncpark (approved):

Issue #140, opened 2023-09-26
Change Kerberos event fields to synchronize with giganto-client 0.13.0


syncpark:
See Kerberos event structure in giganto-client 0.13.0

syncpark:
@dayeon5470 Please modify Kerberos BlockList event fields

PR #141, opened 2023-09-26
Modify kerberos

dayeon5470:
hog: https://github.com/aicers/hog/pull/544
Issue: https://github.com/petabi/review-database/issues/140


minshao:
Migration support is needed for changing `BlockListKerberos` event. Examples could be found in src/migration.rs

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "review-database"
-version = "0.19.1-alpha.2"
+version = "0.19.1-alpha.3"
```
minshao, line 3: Since the introduced change is backward incompatible, this should be 0.20.0-alpha.1, meaning the to-release version would be 0.20.0

minshao (changes requested):

src/migration.rs
```diff
@@ -75,11 +75,18 @@ pub fn migrate_data_dir<P: AsRef<Path>>(data_dir: P, backup_dir: P) -> Result<()
...
-        migrate_0_12_to_0_16,
-    )];
+    let migration: Vec<(_, _, fn(_) -> Result<_, _>)> = vec![
+        (
+            VersionReq::parse(">=0.12.0,<0.16.0")?,
+            Version::parse("0.16.0")?,
+            migrate_0_12_to_0_16,
+        ),
+        (
+            VersionReq::parse(">=0.19.0,<0.20.0")?,
```
minshao, line 85: ">=0.16.0,<0.20.0" here, since version from 0.16.0 to 0.19.0 are all compatible versions the added migration functions can handle.

minshao (approved):

PR #144, opened 2023-10-02
Allow public access for model id in BatchInfo and Scores

minshao:


msk:
Isn't this backward-compatible?

msk (approved):

PR #169, opened 2024-01-11
Added new three detection events

kimhanbeom:
- Add WindowsThreat event for windows sysmon events.
- Add NetworkThreat event for network events.
- Add ExtraThreat event for misc events.


minshao:
1. The `EventKind` enum has been modified to add new variants and remove some existing ones. This breaks the compatibility with the previous versions of the library, so we need to increment the minor version number according to the semantic versioning rules.
2. The `EventMessage` struct, which contains an `EventKind` field, is serialized and stored in the database. To avoid data loss and corruption, we need to implement a migration function that will update the existing records in the database to match the new `EventKind` definition.
3. Please also kindly include the reason why `Log` is removed and replaced with `ExtraThreat` for record keeping.

kimhanbeom:
@minshao 
1.I will fix this.

PR #206, opened 2024-02-07
Add backup create/list fucntions for rocksdb and postgres

kimhanbeom:
This feature is intended for temporary use, driven by project requirements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,23 @@ file is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
 this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- Initial pr
+
+### Added
+
+- Introduced a new module `postgres`. this is module provides the backup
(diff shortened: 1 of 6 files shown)


kimhanbeom:
@msk

Currently, an external project requires backup-related functionality with the following conditions

1. create rocksdb/postgres backup archive. (the created backup archive should be externally transferable and independently available on another server/current server, i.e. rocksdb full backup is required)
2. output a list of backup archives.

Currently, the backup implemented in review-database only supports incremental backup, so it is not possible to create a full backup of a specific point. Therefore, in order to satisfy the above two conditions, I implemented a feature to create a backup through checkpoint rather than backup engine. 

However, in order to use the backup feature of rocksdb more reliably, I think that we should use rocksdb backup engine. That's why I'm thinking of raising an issue/pr for the feature so that the backup engine in the rocks crate I'm currently using supports full backup, and after that feature is added, I'll re-implement the backup feature and include it in main.

For the existing WIP pr, I'm going to close the current pr as I think I just need to reference some reusable features.

msk:
@kimhanbeom, thank you for the detailed update. It sounds good to me.

PR #226, opened 2024-02-15
Avoid exposing filter map directly

minshao:


src/tables/filter.rs
```diff
@@ -0,0 +1,250 @@
...
+
+impl FromKeyValue for Filter {
+    fn from_key_value(key: &[u8], value: &[u8]) -> Result<Self> {
+        use anyhow::anyhow;
+
+        let sep = key
+            .iter()
+            .position(|c| *c == 0)
+            .ok_or(anyhow!("corruptted access token"))?;
+        let username = String::from_utf8_lossy(&key[..sep]).into_owned();
```
msk, line 98: Shouldn't we report an error if the binaries from the database are invalid UTF-8?

msk (approved):

Issue #236, opened 2024-02-27
migration code for `Node`


sophie-cluml:
Make new `Node` have as_is and to_be fields, migrating from current `Node` struct.
Existing field values will be put to `to_be` fields, because no `Node` setting has ever been applied to modules via review-web.
Related to https://github.com/aicers/review-web/pull/129

sophie-cluml:
Please assign this to @BLYKIM

PR #241, opened 2024-03-05
Modify `Node` struct

BLYKIM:
Close: #236


src/migration.rs
```diff
@@ -379,15 +381,17 @@ fn migrate_0_22_to_0_24(store: &super::Store) -> Result<()> {
 
     #[allow(clippy::struct_excessive_bools)]
     #[derive(Deserialize, Serialize)]
-    struct OldNode {
+    pub struct OldNode {
```
minshao, line 384: it's not quite necessary to switch to `pub` here.

minshao (approved):

PR #244, opened 2024-03-06
Migrate Node to have name field

### LINKED ISSUE #243: Do not use `panic!` in `Node` by placing `name` out of `as_is` and `to_be`
In `node.rs`, the `key` calls `panic!` conditionally. This assumes there is no such case, but since it's not 100% guaranteed we need to handle this in a different manner.

sophie-cluml:
closes #243

### DIFF
--- a/src/migration.rs
+++ b/src/migration.rs
@@ -10,7 +10,7 @@ use std::{
     net::IpAddr,
     path::{Path, PathBuf},
 };
-use tracing::info;
+use tracing::{info, warn};
 
 /// The range of versions that use the current database format.
 ///
@@ -32,7 +32,7 @@ use tracing::info;
 /// // the database format won't be changed in the future alpha or beta versions.
 /// const COMPATIBLE_VERSION: &str = ">=0.5.0-alpha.2,<=0.5.0-alpha.4";
 /// ```
-const COMPATIBLE_VERSION_REQ: &str = ">=0.25.0,<=0.26.0-alpha.1";
+const COMPATIBLE_VERSION_REQ: &str = ">0.25.0,<=0.26.0-alpha.2";
 
 /// Migrates data exists in `PostgresQL` to Rocksdb if necessary.
 ///
@@ -184,6 +184,11 @@ pub fn migrate_data_dir<P: AsRef<Path>>(data_dir: P, backup_dir: P) -> Result<()
             Version::parse("0.25.0")?,
             migrate_0_24_to_0_25,
         ),
+        (
+            VersionReq::parse(">=0.25.0,<0.26.0")?,
+            Version::parse("0.26.0")?,
+            migrate_0_25_to_0_26,
+        ),
     ];
 
     let mut store = super::Store::new(data_dir, backup_dir)?;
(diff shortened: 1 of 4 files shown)


src/migration.rs
```diff
@@ -439,13 +598,103 @@ fn migrate_node(store: &super::Store) -> Result<()> {
...
+    #[derive(Deserialize, Serialize)]
+    pub struct NewNode {
+        pub id: u32,
+        pub creation_time: DateTime<Utc>,
+        pub as_is: Option<NewNodeSetting>,
+        pub to_be: Option<NewNodeSetting>,
+    }
+
+    impl FromKeyValue for NewNode {
+        // pita?
```
sehkone, line 663: Please, remove this.

sehkone:
Given that a `name` can be changed and we need to save user input, I propose adding an extra field to the `Node`. Furthermore, the current name, `to_be`, might be misinterpreted as the immediate next value, which isn't always the case. To address this, I recommend using the `_draft` prefix. With this change, the `Node` would be structured as follows:

```
pub struct Node {
    pub id: u32,
    pub name: String,
    pub name_draft: Option<String>,
    pub setting: Option<NodeSetting>,
    pub setting_draft: Option<NodeSetting>,
    pub creation_time: DateTime<Utc>,
}
```

src/migration.rs
```diff
@@ -32,7 +32,7 @@ use tracing::info;
 /// // the database format won't be changed in the future alpha or beta versions.
 /// const COMPATIBLE_VERSION: &str = ">=0.5.0-alpha.2,<=0.5.0-alpha.4";
 /// ```
-const COMPATIBLE_VERSION_REQ: &str = ">=0.25.0,<0.26.0-alpha";
+const COMPATIBLE_VERSION_REQ: &str = ">=0.25.0,<=0.26.0-alpha.1";
```
minshao, line 35: This should be ">0.25.0,<=0.26.0-alpha.1", probably my last mistake have caused the confusion. My apologies.

sophie-cluml:
@sehkone, for`name_draft` field, will the value in the `name_draft` moved to `name` when "apply" is completed? Just like how `as_is` and `to_be` field behave?
I am confused a bit. What I understood as your direction from our last google meet was to have `name` field only. I'd like to be in sync, so please guide me which way I should take.

By the way, I think it's okay either way, to have `name_draft` field, and not to have it. 
- If we have `name_draft`, then change of `name` will be tied to "apply". (To me it makes sense, because it aligns with  "apply" lifecycle, except for the initial insertion)
- Meanwhile, if we do not have `name_draft`, then change of `name` will be directly saved to `name`, regardless of "apply". (`name` is not `hostname` of node, but just a human-readable name of an item in db, so it makes sense to me to disintegrate `name` change and "apply".)

sehkone:
There are two things we should take care of.

- When a `Node` is initially created, the creation method must receive a non-null name as an input argument and store it in the `name` of the `Node`.
- When the apply method is called, if the `name_draft` is provided with a non-null value, then the `name` should be updated to this value.

There are no other cases when the `name` is changed except the above.

src/migration.rs
sophie-cluml: done

src/migration.rs
sophie-cluml: done

sehkone (approved):

src/migration.rs
```diff
@@ -259,6 +264,164 @@ fn read_version_file(path: &Path) -> Result<Version> {
...
+        }
+    }
+
+    impl From<OldNode> for Node {
+        fn from(input: OldNode) -> Self {
+            let (name, name_draft) = match (&input.as_is, &input.to_be) {
+                (None, None) => {
+                    warn!("Both as_is and to_be is None. So random name will be set for the Node.");
+                    let random_as_is_name = format!("random-node-name-{}", rand::random::<u64>());
+                    (random_as_is_name, None)
```
sehkone, line 394: In the unlikely event of a duplicate, the migration will stop. This condition is unlikely to occur in the first place, so I'd rather not deal with it.

sophie-cluml: Great idea. Done.

minshao (approved):

PR #261, opened 2024-03-20
Avoid exposing tidb map

minshao:


src/tables/tidb.rs
```diff
@@ -0,0 +1,287 @@
...
+    pub id: u32,
+    pub name: String,
+    pub description: Option<String>,
+    pub kind: Kind,
+    pub version: String,
+    pub patterns: Vec<Rule>,
+}
+
+impl Tidb {
+    /// Parse and validate input TI database
```
msk, line 24: * Parse → Parses
* validate → validates

Check other documentations as well.

msk (changes requested):

msk (approved):

PR #270, opened 2024-04-11
Add new LockyRansomware detection event

### LINKED ISSUE #268: Add new detection event for Locky Ransomware semi-supervised model
Related issue aicers/review-web#197

syncpark:
Add new detection event: `LockyRansomware`

Close #268

### DIFF
--- a/src/event/dns.rs
+++ b/src/event/dns.rs
@@ -1,5 +1,5 @@
 #![allow(clippy::module_name_repetitions, clippy::struct_excessive_bools)]
-use super::{common::Match, EventCategory, TriagePolicy, TriageScore, MEDIUM};
+use super::{common::Match, EventCategory, TriagePolicy, TriageScore, HIGH, MEDIUM};
 use chrono::{serde::ts_nanoseconds, DateTime, Local, Utc};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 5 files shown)


src/event/dns.rs
```diff
@@ -136,7 +136,7 @@ impl Match for DnsCovertChannel {
     }
 
     fn level(&self) -> NonZeroU8 {
-        MEDIUM
+        HIGH
```
msk, line 139: This changes `DnsCovertChannel`. Since this PR is to add `LockyRansomware`, shouldn't this change be in a separate PR, with proper justification?

syncpark: Do you think the `LockRansomware` event struct should contain multiple protocol events like Conn or Http instead of only Dns?

Currently `LockyRansomware` model in Hog is learned and created only from Dns events. And the `LockyRansomware` detection event uses same event fields with `DnsCovertChannel`.

If `LockyRansomware` model can be created from various protocol events, a new struct is needed to describe the event.

```rust
struct LockyRansomware {
   service: String, // "Dns"
   fields: Vec<u8>, // Dns event struct of Giganto if service is "Dns"
   confidence: f32
}
```
The `fields` can have type `String` if the `fields` contains JSON string instead of serialized struct.

msk (approved):

PR #278, opened 2024-05-15
Add `Table<AccessToken>::tokens`

minshao:


src/tables/access_token.rs
```diff
@@ -73,4 +75,49 @@ impl<'d> Table<'d, AccessToken> {
         let (key, _value) = AccessToken::create_key_value(username, token);
         self.map.get(&key).map(|v| v.is_some())
     }
+
+    /// Find all `token`s for `username` in the database.
```
msk, line 79: "Find" → "Finds"

src/tables/access_token.rs
```diff
@@ -73,4 +75,49 @@ impl<'d> Table<'d, AccessToken> {
         let (key, _value) = AccessToken::create_key_value(username, token);
         self.map.get(&key).map(|v| v.is_some())
     }
+
+    /// Finds all `token`s for `username` in the database.
```
msk, line 79: Since "token" is not an identifier in the code, I don't think it needs enclosing backticks.

msk (approved):

Issue #284, opened 2024-05-29
Modify `Event` enum and add migration code


kimhanbeom:
## Background

- We are working on updating the gigatno-client version to `0.17.0` for all modules that use giganto-client.
- In giganto-client `0.17.0`, fields in the SSH, HTTP, SMTP, CONN, NTLM, and TLS protocols have changed, and REview that receive detected events from these protocols also need to change the field values in `Event` and add migration code.

## Detail task

- Modify the fields in `Event` corresponding to SSH, HTTP, SMTP, CONN, NTLM, and TLS protocols.
- Add migration functionality based on changes to the `Event` field.

kimhanbeom:
I'll take care of the current issue.

sophie-cluml:
[TEST] I've been asked to comment for testing github comment and assignee option, so we're commenting for a while. We'll remove it after testing shortly. Sorry for inconvinience.

PR #285, opened 2024-05-30
Modify event struct and add migration code

kimhanbeom:
Close: #284


src/event/http.rs
```diff
@@ -140,6 +140,12 @@ pub struct HttpThreatFields {
     pub content_encoding: String,
     pub content_type: String,
     pub cache_control: String,
+    pub orig_filenames: Vec<String>,
+    pub orig_mime_types: Vec<String>,
+    pub resp_filenames: Vec<String>,
+    pub resp_mime_types: Vec<String>,
+    pub post_body: Vec<u8>,
+    pub state: String,
```
minshao, line 148: Is it possible to use `enum` or status code to represent `state` instead of `String`?

kimhanbeom: Yes, it can be changed to an enum type. However, since other repositories are already using `state` of type `string`, I will open a new issue to change it to an enum type in the future.

sehkone: @kimhanbeom, Would you create the issue even if we address it sometime later?

minshao: Got it. In addition, IMHO, the same should apply to fields like `resp_mime_types`, `conn_state`, etc. it would be great if it's mentioned in the issue.

kimhanbeom: I've opened an issue about this.

minshao (approved):

PR #295, opened 2024-06-20
Add `status` for `Agent`

minshao:


sehkone:
As we discussed [here](https://github.com/petabi/review-database/discussions/294#discussioncomment-9834024), I agree with your suggestion to use an `enum` for the `Status`.

sehkone (approved):

PR #296, opened 2024-06-21
Modify display message format of EventMessage and Event

### LINKED ISSUE #274: Send all fields of detected events to Syslog
Related to: https://github.com/aicers/patio/issues/45, https://github.com/aicers/patio/issues/85

- [x]  All fields of detected events should be sent to Syslog
- [x]  The delimiter (currently comma(,)) should not be included in the field content. The comma delimiter may need to be changed.

syncpark:
Close #274 

Modify the display message format of `Event` and `EventMessage` structure.
The modified messages will be sent to syslog according to RFC5424.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -23,7 +23,7 @@ mod tor;
 use std::{
     collections::HashMap,
     convert::TryInto,
-    fmt,
+    fmt::{self},
     net::IpAddr,
     num::NonZeroU8,
     sync::{Arc, Mutex, MutexGuard},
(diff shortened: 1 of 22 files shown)


src/event.rs
```diff
@@ -111,6 +111,14 @@ const NETWORK_THREAT_EVENT: &str = "Network Threat Events";
 const MISC_LOG_THREAT: &str = "Log Threat";
 const LOCKY_RANSOMWARE: &str = "Locky Ransomware";
 
+const PRIVATE_ENTERPRISE_NUMBER: &str = "000000";
```
minshao, line 114: Since this number should be enterprise or customer specific, I think it's better not to be hard coded.

src/event.rs
```diff
@@ -111,6 +111,14 @@ const NETWORK_THREAT_EVENT: &str = "Network Threat Events";
...
+
+// Message ID for the detected events
+const MSGID: &str = "DETECT";
```
minshao, line 117: Any reason to hard code `MSGID`?

src/event.rs
```diff
@@ -111,6 +111,14 @@ const NETWORK_THREAT_EVENT: &str = "Network Threat Events";
...
+
+// RFC5424 Structured Data: (MSGID, STRUCTURED-DATA, MSG)
+type OrderedStructuredData = (String, HashMap<String, Vec<(String, String)>>, String);
```
minshao, line 120: * `OrderedStructuredData::0` is hard coded to `MSGID`, so I think it's a redundant element.
* `Ordered` gives the impression that the type is following a certain order. However, since it's a tuple, the order among the elements is fixed by nature, therefore, IMHO, it's not necessary.
* Can you explain what is "STRUCTURED-DATA"?

src/event.rs
syncpark: The PEN (Private Enterprise Number) is the registered number of our company, not for customer.

PEN typically used in LDAP and SNMP MIB configurations.
Our company must be registered via the following URL to get the PEN.

https://www.iana.org/assignments/enterprise-numbers/assignment/apply/

Until we register our company PEN, I think it's better to use "000000" instead.

I moved it inside `generate_sdid()` function.

src/event.rs
syncpark: RFC5424 defines syslog message format like follow:
```
  SYSLOG-MSG = HEADER SP STRUCTURED-DATA [SP MSG]
  HEADER = ... MSGID
  STRUCTURED-DATA = NILVALUE / 1*SD-ELEMENT
  SD-ELEMENT = "[" SD-ID *(SP SD-PARAM) "]"
```

The `OrderedStructuredData` type will be used in the `syslog` crate to write syslog messages.

I define the MSGID for the detected events to `DETECT`. Later, I think that the MSGID `OPLOG` or `OUTLIER` will be needed to identify the operation log or outlier events.

The `OrderedStructuredData` is named and defined to replace `StructuredData` of `syslog` crate.
In `syslog` crate, `StructuredData` is defined to `HashMap<String, HashMap<String, String>>`.

SD-PARAM is the key-value pair of the detected events. Using the `StructuredData` type, the key-values pairs will be written into syslog file without order, because the crate defines SD-PARAM to `HashMap<String, String>`.

I think that maintaining the order of SD-PARAMs makes it easier for engineers to read and understand. To resolve this issue, I implement a new LogFormat for SD-PARAM to use `Vec<(String, String)>`.

In the PR petabi/review#1616 use this modification.

src/event.rs
syncpark: MSGID is used to describe the types of syslog message.

RFC5424 explains that MSDID is used for like following:
```
For example, a firewall might use the MSGID "TCPIN" for incoming TCP traffic and the MSGID "TCPOUT" for outgoing TCP traffic
```

I think the MSGID will be extended to ENUM values in the near future to include "OPLOG" or "OUTLIER" to identify the message type "operation log" or "outlier event".

But currently only `DETECT` MSGID is used, I removed the `const` value.

src/event.rs
msk: @syncpark, I understand your rationale behind using the placeholder value "000000" for the Private Enterprise Number (PEN) until we register our company's PEN. However, I have some concerns about hardcoding this value, even as a temporary measure:

1. **RFC 5424 Compliance**: Hardcoding a placeholder value like "000000" doesn't conform to the requirements of RFC 5424. The PEN is intended to be a unique identifier, and using an incorrect or placeholder value could lead to non-compliance and potential issues with interoperability.

2. **Library vs. Application Responsibility**: The determination of the PEN should be the responsibility of the application using this crate, not the crate itself. The review-database crate should be designed to accept a PEN provided by the application, allowing for greater flexibility and ensuring that the correct PEN is used in various deployment environments.

To address these concerns, I suggest refactoring the code to allow the PEN to be passed as a parameter to review-database. This way, the application using the review-database crate can provide the appropriate PEN, and we can ensure compliance with RFC 5424 without hardcoding any values.

src/event.rs
minshao: I acknowledge that maintaining a specific order for SD-PARAMs enhances readability. However, switching to Vec<_> poses significant challenges in ensuring the uniqueness of parameter names within the scope. Moreover, without comprehensive documentation, maintaining a consistent fixed order could be problematic.

Therefore, I propose retaining the use of HashMap<_, _>, which inherently handles uniqueness. To address the readability concern, we can implement a mechanism to print the parameters in the user’s or engineer’s preferred order. Since I observe that different events do share fields that are preferred to be listed in the same order, a sorting helper function could be provided as a convenience method to facilitate ordered output for different events. This approach balances readability with the inherent advantages of HashMap<_, _> in maintaining distinct parameter names.

sehkone:
@syncpark 
As you know, we have a rule against public repositories linking to locations of private repositories, even if only authorized personnel can access them. Could you remove these links?

minshao (approved):

PR #305, opened 2024-07-08
Moved EventCategory definition to review-protocol

### LINKED ISSUE #304: Move EventCategory definition to review-protocol crate
https://github.com/petabi/review-database/blob/c52ee92e4393026cfd5fcbfaa9edab88795cf0d9/src/types.rs#L44-L55

We've decided to assign EventCategory to the detected events when they are created.
To share the EventCategory definition between server and client(analyzer), the definition must be moved to `review-protocol` crate.

- [x] Move `EventCategory` definition to `review-protocol` crate
- [x] Remove the `HttpThreat` member from the EventCategory
- [x] Modify related code

syncpark:
Close #304 
- Moved `EventCategory` definition to review-protocol crate
- `HttpThreat` category is changed to **obsoleted** and replaced existing `HttpThreat` category value to `Reconnaissance`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   Modified messages will be sent to syslog.
 - Used `FromPrimitive` and `ToPrimitive` for converting `EventCategory` instead of
   manually implementing `TryFrom`.
+- `EventCategory` definition is moved to `review-protocol`.
 
 ### Removed
 
@@ -51,6 +52,12 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Corrected `EventCategory` for events:
   - `RdpBruteForce`: Changed from `Exfiltration` to `Discovery`.
   - `HttpThreat`: Changed from `HttpThreat` to `Reconnaissance`.
+
+### Deprecated
+
+- `EventCategory::HttpThreat` is deprecated and replaced with `Reconnaissance`.
+  After the modification of all programs using `EventCategory::HttpThreat`
+  are completed, the `HttpThreat` category will be deleted.
 
 ## [0.28.0] - 2024-05-16
 
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,3 +1,4 @@
+#![allow(deprecated)]
 use std::{cmp::Ordering, net::IpAddr, ops::RangeInclusive};
 
 use anyhow::Result;
@@ -42,6 +43,7 @@ pub struct PretrainedModel(pub Vec<u8>);
 )]
 #[repr(u8)]
 pub enum EventCategory {
+    Unknown = 0,
     Reconnaissance = 1,
     InitialAccess,
     Execution,
@@ -51,6 +53,7 @@ pub enum EventCategory {
     CommandAndControl,
     Exfiltration,
     Impact,
+    #[deprecated(note = "Users should use `EventCategory::Reconnaissance` instead")]
     HttpThreat,
 }
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -78,8 +78,8 @@ pub use self::{
     tor::{TorConnection, TorConnectionFields},
 };
 use super::{
-    types::{Endpoint, EventCategory, HostNetworkGroup},
-    Customer, Network, TriagePolicy,
+    types::{Endpoint, HostNetworkGroup},
+    Customer, EventCategory, Network, TriagePolicy,
 };
 
 // event levels (currently unused ones commented out)
@@ -1675,8 +1675,6 @@ impl From<&EventKind> for EventCategory {
             | EventKind::NonBrowser
             | EventKind::CryptocurrencyMiningPool => EventCategory::CommandAndControl,
 
-            EventKind::HttpThreat => EventCategory::HttpThreat,
-
             EventKind::RdpBruteForce => EventCategory::Discovery,
             EventKind::RepeatedHttpSessions => EventCategory::Exfiltration,
 
@@ -1699,7 +1697,8 @@ impl From<&EventKind> for EventCategory {
             | EventKind::BlockListSsh
             | EventKind::BlockListTls => EventCategory::InitialAccess,
 
-            EventKind::PortScan
+            EventKind::HttpThreat
+            | EventKind::PortScan
             | EventKind::MultiHostPortScan
             | EventKind::NetworkThreat
             | EventKind::ExtraThreat => EventCategory::Reconnaissance,
@@ -1869,7 +1868,7 @@ impl fmt::Display for EventMessage {
             "time={:?} event_kind={:?} category={:?} ",
             self.time.to_rfc3339(),
             format!("{:?}", self.kind),
-            EventCategory::from(&self.kind).to_string()
+            format!("{:?}", EventCategory::from(&self.kind))
         )?;
         let _r = match self.kind {
             EventKind::DnsCovertChannel => bincode::deserialize::<DnsEventFields>(&self.fields)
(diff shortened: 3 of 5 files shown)


src/event/conn.rs
```diff
@@ -7,8 +7,11 @@ use std::{
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 
-use super::{common::Match, EventCategory, TriagePolicy, TriageScore, MEDIUM};
-use crate::event::common::{triage_scores_to_string, vector_to_string};
+use super::{common::Match, TriagePolicy, TriageScore, MEDIUM};
+use crate::{
+    event::common::{triage_scores_to_string, vector_to_string},
+    types::EventCategory,
```
msk, line 13: The `review-database` crate still retains its own `EventCategory` enum. I'm not clear on why all usages of it are being replaced with `review_protocol::EventCategory`. Directly using `review_protocol::EventCategory` throughout `review-database` could potentially create unnecessary coupling between the two crates.

Could you explain the reasoning behind using `review_protocol::EventCategory` instead of `review_database::EventCategory` within the review-database codebase? What are the advantages you see in this approach?

msk:
@minshao Could you check if changing the definition of `enum EventCategory` requires any database migration? Other than that, I’m good with this PR.

minshao:
> @minshao Could you check if changing the definition of `enum EventCategory` requires any database migration? Other than that, I’m good with this PR.

Since  `EventCategory::HttpThreat` is removed, existing database entries contains `EventCategory::HttpThreat` needs to be updated. That is, all entries in `TriagePolicy` map needs to be migrated.

minshao:
I'd like to clarify the intention behind the recent changes to the `HttpThreat` field. 

As a reminder, database migration is typically only necessary when there are changes to the underlying database fields. In the case of deprecation, where a field is no longer used or is being replaced, migration is usually not required.

However, in the latest commit (commit ID: 4332d9989c10c45a00440d2166175dc4c4c91c42):
* the CHANGELOG states that `HttpThreat` **is removed**. 
* Upon reviewing the code, I noticed that `HttpThreat` is only marked as **"obsoleted"** and **not** actually removed. 
* Furthermore, the migration code is updating existing entries to **replace** `HttpThreat` with `Reconnaissance`.

I'm having trouble understanding the intended outcome here. @syncpark Should we be removing HttpThreat entirely, or is it intended to remain as an obsolete field? I'd appreciate any clarification on this matter to ensure a smooth transition.

syncpark:
> I'd like to clarify the intention behind the recent changes to the `HttpThreat` field.
> 
> As a reminder, database migration is typically only necessary when there are changes to the underlying database fields. In the case of deprecation, where a field is no longer used or is being replaced, migration is usually not required.
> 
> However, in the latest commit (commit ID: [4332d99](https://github.com/petabi/review-database/commit/4332d9989c10c45a00440d2166175dc4c4c91c42)):
> 
> * the CHANGELOG states that `HttpThreat` **is removed**.
> * Upon reviewing the code, I noticed that `HttpThreat` is only marked as **"obsoleted"** and **not** actually removed.
> * Furthermore, the migration code is updating existing entries to **replace** `HttpThreat` with `Reconnaissance`.
> 
> I'm having trouble understanding the intended outcome here. @syncpark Should we be removing HttpThreat entirely, or is it intended to remain as an obsolete field? I'd appreciate any clarification on this matter to ensure a smooth transition.

I'm sorry for causing the confusion.

The initial intention was to delete it. However, it seemed like it would be a problem to delete the enum value while the programs that use that value had not yet been modified, so I modified it with **obsolete** while writing the migration code. Deleting this value will be done after all program modifications have been completed.

The `HttpThreat` category will be removed in the near future, all the existing `HttpThreat` category values are replaced to `Reconnaissance`. And a new field `category` that is inserted in all the detected event structures will be used instead of  hardcoded category value.

I modified the CHANGELOG.

minshao:
> I'm sorry for causing the confusion.
> 
> The initial intention was to delete it. However, it seemed like it would be a problem to delete the enum value while the programs that use that value had not yet been modified, so I modified it with **obsolete** while writing the migration code. Deleting this value will be done after all program modifications have been completed.
> 
> The `HttpThreat` category will be removed in the near future, all the existing `HttpThreat` category values are replaced to `Reconnaissance`. And a new field `category` that is inserted in all the detected event structures will be used instead of hardcoded category value.
> 
> I modified the CHANGELOG.

No worries, thanks for the clarification, that's very helpful. I'll proceed accordingly.

src/types.rs
```diff
@@ -51,7 +52,7 @@ pub enum EventCategory {
     CommandAndControl,
     Exfiltration,
     Impact,
-    HttpThreat,
+    HttpThreat, // Obsoleted. Do not use this category
```
minshao, line 55: Instead of comment, [deprecated attribute](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute) could be more helpful.
```suggestion
    #[deprecated(note = "Users should use `EventCategory::Reconnaissance` instead")]
    HttpThreat, 
```

CHANGELOG.md
```diff
@@ -33,6 +33,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   Modified messages will be sent to syslog.
 - Used `FromPrimitive` and `ToPrimitive` for converting `EventCategory` instead of
   manually implementing `TryFrom`.
+- `EventCategory` definition is moved to `review-protocol`.
+  `HttpThreat` category is obsoleted and replaced to `Reconnaissance`.
```
minshao, line 37: * Since `HttpThreat` is only deprecated at this stage, it should be mentioned in deprecated section in CHANGELOG, see line 315 of CHANGELOG.
* "replaced to" -> "replaced with"?

src/migration.rs
minshao, line 1: Since `HttpThreat` is not removed yet and is still part of the Public API, it is not necessary to migrate all existing data entries at this point.

minshao (approved):

src/types.rs
minshao: After addressing the clippy warnings, it looks good to me.

syncpark: With the directive `#[deprecated(note = "...")]`, `HttpThreat` enum member always makes clippy error like following:
```rust
error: use of deprecated unit variant `types::EventCategory::HttpThreat`: Users should use `EventCategory::Reconnaissance` instead
  --> src/types.rs:56:5
   |
56 |     HttpThreat, // Obsoleted. Do not use this category
   |     ^^^^^^^^^^
   |
   = note: `-D deprecated` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(deprecated)]`
```

The directive to indicate that it is `deprecated` creates a deprecated clippy error. I haven't found a solution yet.

I changed it to comment now.

minshao: Sorry, I didn't realize that proc macro would still cause the warning even with `#[allow(deprecated)]`. The following should do the trick.
```diff
diff --git a/src/types.rs b/src/types.rs
index 0a7cb57..0b41480 100644
--- a/src/types.rs
diff --git a/src/types.rs b/src/types.rs
index 0a7cb57..0b41480 100644
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,3 +1,4 @@
+#![allow(deprecated)]
 use std::{cmp::Ordering, net::IpAddr, ops::RangeInclusive};
 
 use anyhow::Result;
@@ -52,7 +53,8 @@ pub enum EventCategory {
     CommandAndControl,
     Exfiltration,
     Impact,
-    HttpThreat, // Obsoleted. Do not use this category
+    #[deprecated(note = "Users should use `EventCategory::Reconnaissance` instead")]
+    HttpThreat,
 }
 
 #[derive(Deserialize)]
```

minshao (approved):

Issue #306, opened 2024-07-09
Migrate `Agent::config` to have full configuration


sophie-cluml:
Currently, only a few of the individual agent's configuration items are stored in the review-database. However, we want to move towards storing full configuration in the review-database. In order to move in this direction, we want to write code that migrates the current `Agent::config` to have full configuration items. This issue does not aim to change the type of `Agent::config`. We see the targets of this migration as `node_map`, due to Giganto, and `agents_map`, due to agents other than Giganto.

sehkone:
@sophie-cluml,
I don’t think the "more configuration items" you described align exactly with what we discussed. The configuration items should comprehensively cover everything each agent needs since we require every agent to fetch its complete configuration from the central server.

sophie-cluml:
Thank you for the fix!

PR #307, opened 2024-07-11
Migrate `Agent::config` to have full configuration

BLYKIM:
Closes: #306

### DIFF
--- a/src/migration.rs
+++ b/src/migration.rs
@@ -6,8 +6,9 @@ mod migration_structures;
 use std::{
     fs::{create_dir_all, File},
     io::{Read, Write},
-    net::IpAddr,
+    net::{IpAddr, Ipv4Addr, SocketAddr},
     path::{Path, PathBuf},
+    str::FromStr,
 };
 
 use anyhow::{anyhow, Context, Result};
@@ -37,7 +38,7 @@ use crate::{Agent, AgentStatus, Giganto, Indexed, IterableMap};
 /// // the database format won't be changed in the future alpha or beta versions.
 /// const COMPATIBLE_VERSION: &str = ">=0.5.0-alpha.2,<=0.5.0-alpha.4";
 /// ```
-const COMPATIBLE_VERSION_REQ: &str = ">=0.29.0-alpha.6,<=0.29.0-alpha.6";
+const COMPATIBLE_VERSION_REQ: &str = ">=0.29.0-alpha.7,<=0.29.0-alpha.7";
 
 /// Migrates data exists in `PostgresQL` to Rocksdb if necessary.
 ///
@@ -363,105 +364,6 @@ fn migrate_0_29_node(store: &super::Store) -> Result<()> {
         pub sensors: bool,
         pub sensor_list: HashMap<String, bool>,
     }
-    #[allow(clippy::struct_excessive_bools)]
-    #[derive(Clone, Default, Deserialize, Serialize, PartialEq)]
-    pub struct NodeSettings {
-        pub customer_id: u32,
-        pub description: String,
-        pub hostname: String,
-
-        pub piglet: bool,
-        pub piglet_giganto_ip: Option<IpAddr>,
-        pub piglet_giganto_port: Option<PortNumber>,
-        pub save_packets: bool,
-        pub http: bool,
-        pub office: bool,
-        pub exe: bool,
-        pub pdf: bool,
-        pub vbs: bool,
-        pub txt: bool,
-        pub smtp_eml: bool,
-        pub ftp: bool,
-
-        pub giganto: bool,
-        pub giganto_ingestion_ip: Option<IpAddr>,
-        pub giganto_ingestion_port: Option<PortNumber>,
-        pub giganto_publish_ip: Option<IpAddr>,
-        pub giganto_publish_port: Option<PortNumber>,
-        pub giganto_graphql_ip: Option<IpAddr>,
-        pub giganto_graphql_port: Option<PortNumber>,
-        pub retention_period: Option<u16>,
-
-        pub reconverge: bool,
-
-        pub hog: bool,
-        pub hog_giganto_ip: Option<IpAddr>,
-        pub hog_giganto_port: Option<PortNumber>,
-        pub protocols: Option<Vec<String>>,
-
-        pub sensors: Option<Vec<String>>,
-    }
-    impl From<OldNodeSettings> for NodeSettings {
-        fn from(input: OldNodeSettings) -> Self {
-            let txt = input.html || input.txt;
-
-            let protocols = if input.protocol_list.is_empty() {
-                None
-            } else {
-                let mut list: Vec<String> = input
-                    .protocol_list
-                    .into_iter()
-                    .filter_map(|(k, v)| if v { Some(k) } else { None })
-                    .collect();
-                list.sort_unstable();
-                Some(list)
-            };
-
-            let sensors = if input.sensor_list.is_empty() {
-                None
-            } else {
-                let mut list: Vec<String> = input
-                    .sensor_list
-                    .into_iter()
-                    .filter_map(|(k, v)| if v { Some(k) } else { None })
-                    .collect();
-                list.sort_unstable();
-                Some(list)
-            };
-
-            Self {
-                customer_id: input.customer_id,
-                description: input.description,
-                hostname: input.hostname,
-                piglet: input.piglet,
-                piglet_giganto_ip: input.piglet_giganto_ip,
-                piglet_giganto_port: input.piglet_giganto_port,
-                save_packets: input.save_packets,
-                http: input.http,
-                office: input.office,
-                exe: input.exe,
-                pdf: input.pdf,
-                txt,
-                vbs: false,
-                smtp_eml: input.smtp_eml,
-                ftp: input.ftp,
-                giganto: input.giganto,
-                giganto_ingestion_ip: input.giganto_ingestion_ip,
-                giganto_ingestion_port: input.giganto_ingestion_port,
-                giganto_publish_ip: input.giganto_publish_ip,
-                giganto_publish_port: input.giganto_publish_port,
-                giganto_graphql_ip: input.giganto_graphql_ip,
-                giganto_graphql_port: input.giganto_graphql_port,
-                retention_period: input.retention_period,
-                reconverge: input.reconverge,
-                hog: input.hog,
-                hog_giganto_ip: input.hog_giganto_ip,
-                hog_giganto_port: input.hog_giganto_port,
-                protocols,
-                sensors,
-            }
-        }
-    }
 
     impl From<OldNodeSettings> for NodeProfile {
         fn from(input: OldNodeSettings) -> Self {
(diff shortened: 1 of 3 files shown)


src/migration/migration_structures.rs
```diff
@@ -509,3 +544,206 @@ impl From<BlockListTlsBeforeV29> for BlockListTlsFields {
...
+
+    pub dump_dir: String,
+
+    pub dump_items: Option<Vec<DumpItem>>,
+    pub dump_http_content_types: Option<Vec<DumpHttpContentType>>,
+
+    pub giganto_ingest_srv_addr: SocketAddr,
+    pub giganto_name: String,
+
+    pub pcap_max_size: u32, // 단위: MB
```
sehkone, line 566: Writing in English is preferred.

src/migration/migration_structures.rs
```diff
@@ -1,14 +1,49 @@
...
+pub const DEFAULT_GIGANTO_EXPORT_DIR: &str = "/data/datalake/export";
+pub const DEFAULT_RETENTION: u16 = 100;
+pub const DEFAULT_ACK_TRANSMISSION: u16 = 1024;
+pub const DEFAULT_MAX_OPEN_FILES: i32 = 8000;
+pub const DEFAULT_MAX_MB_OF_LEVEL_BASE: u64 = 512;
+pub const DEFAULT_NUM_OF_THREADS: i32 = 8;
+pub const DEFAULT_MAX_SUB_COMPACTION: u32 = 2;
+
+// common
+pub const DEFAULT_LOG_DIR: &str = "/data/logs/apps";
```
sehkone, line 45: @sophie-cluml @BLYKIM,
This part seems to divulge agents' internal information. I think it would suffice for this migration to transfer the old data into the new format without adding any other information, even if it does not guarantee the agents' full operation.

src/migration.rs
```diff
@@ -477,63 +379,47 @@ fn migrate_0_29_node(store: &super::Store) -> Result<()> {
...
                         },
-                        sensors: if s.sensors {
+                        active_sources: if s.sensors {
                             let mut list: Vec<_> = s.sensor_list.keys().cloned().collect();
                             list.sort_unstable();
                             Some(list)
                         } else {
                             None
                         },
+                        active_models: None,
```
sehkone, line 412: I heard that missing fields in TOML are deserialized to `None`. If so, this would be unnecessary. Could you check this out?

src/migration.rs
```diff
@@ -477,63 +379,47 @@ fn migrate_0_29_node(store: &super::Store) -> Result<()> {
...
+                        giganto_publish_srv_addr: s.hog_giganto_ip.map(|ip| {
+                            let port = s.hog_giganto_port.unwrap_or(38371);
+                            SocketAddr::new(ip, port)
+                        }),
+                        giganto_name: None,
+                        cryptocurrency_mining_pool: "/usr/local/aice/tidb/cryptocurrency.json"
+                            .to_string(),
+                        log_dir: "/data/logs/apps".to_string(),
+                        export_dir: "/data/export".to_string(),
+                        services_path: "/usr/local/aice/data/services".to_string(),
```
sehkone, line 422: Since default values can be handled in both the AI and each agent, I think the empty string would suffice.

sehkone: I suggest this approach in other similar contexts.

src/migration.rs
sehkone: I suggest applying this approach to other similar cases.

BLYKIM: Then, are you suggesting to remove the fields that are initialized to `None` from each agent's config field?

sehkone: Yes. I don't think we should disclose any unnecessary information more than absolutely necessary.

src/migration.rs
BLYKIM: > Since default values can be handled in both the AI and each agent

Should I replace all instances of fixed default strings in this commit with empty strings?

sehkone: That's exactly what I suggested.

src/migration/migration_structures.rs
BLYKIM: done

src/migration.rs
```diff
@@ -477,63 +379,44 @@ fn migrate_0_29_node(store: &super::Store) -> Result<()> {
...
                         } else {
                             None
                         },
-                        sensors: if s.sensors {
+                        active_sources: if s.sensors {
                             let mut list: Vec<_> = s.sensor_list.keys().cloned().collect();
                             list.sort_unstable();
                             Some(list)
                         } else {
                             None
```
sophie-cluml, line 410: I also have an understanding that `active_models` takes the same approach as `active_protocols`, so I think the `else` block needs to return `Some` of empty vector.

src/migration/migration_structures.rs
BLYKIM: I replaced string values with empty string, but for numeric values, I used fixed in the code without defining constants.

src/migration.rs
BLYKIM: done

src/migration.rs
BLYKIM: done

src/migration.rs
```diff
@@ -548,17 +431,65 @@ fn migrate_0_29_node(store: &super::Store) -> Result<()> {
...
+                            }
+                            if s.smtp_eml {
+                                list.push(DumpItem::Eml);
+                            }
+                            if s.ftp {
+                                list.push(DumpItem::Ftp);
+                            }
+
+                            if list.is_empty() {
+                                None
```
sophie-cluml, line 458: From my understanding, "Features" module will dump all items when `dump_items == None`, so I think this block needs to return `Some(vec![])`.

src/migration.rs
```diff
@@ -548,17 +431,65 @@ fn migrate_0_29_node(store: &super::Store) -> Result<()> {
...
+                            }
+                            if s.exe {
+                                list.push(DumpHttpContentType::Exe);
+                            }
+                            if s.pdf {
+                                list.push(DumpHttpContentType::Pdf);
+                            }
+
+                            if list.is_empty() {
+                                None
```
sophie-cluml, line 479: I think we are taking the same approach to `dump_http_content_types` as `dump_items`, so I think this block needs to return `Some(vec![])`.

src/migration.rs
```diff
@@ -477,63 +379,53 @@ fn migrate_0_29_node(store: &super::Store) -> Result<()> {
...
                         },
-                        sensors: if s.sensors {
+                        active_sources: if s.sensors {
                             let mut list: Vec<_> = s.sensor_list.keys().cloned().collect();
                             list.sort_unstable();
                             Some(list)
                         } else {
                             None
                         },
+                        active_models: None,
```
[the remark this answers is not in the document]
BLYKIM: I understand the difference between `None` and default values. So, I'm curious if it's okay to use `Some(DEFAULT_MODEL_LIST)` in this code.

@sophie-cluml @sehkone

src/migration.rs
BLYKIM: I think it was `active_sources` not `active_models`. I'll fix it!

src/migration.rs
```diff
@@ -477,63 +379,47 @@ fn migrate_0_29_node(store: &super::Store) -> Result<()> {
...
                             list.sort_unstable();
-                            Some(list)
+
+                            let res_list = list
+                                .into_iter()
+                                .filter_map(|s| ProtocolForHog::from_str(&s).ok())
+                                .collect();
+                            Some(res_list)
                         } else {
                             None
```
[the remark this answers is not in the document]
BLYKIM: done

src/migration.rs
BLYKIM: done

src/migration.rs
BLYKIM: done

src/migration.rs
BLYKIM: done

src/migration.rs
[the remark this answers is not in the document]
sophie-cluml: I think `Some(vec![])` would be slightly preferred because what I understood from this comment by Kone,

> Yes. I don't think we should disclose any unnecessary information more than absolutely necessary.

is that it seems more important, not to disclose info, and I think that includes default model list.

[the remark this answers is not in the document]
sophie-cluml: I didn't see Kone's comment when I wrote my comment https://github.com/petabi/review-database/pull/307#discussion_r1673778842, so I'd appreciate if you could consider above comment as my most recent idea.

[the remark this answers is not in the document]
sophie-cluml: By the way, I think `None` for this one maybe needs care, because I have been told that a few models cause heavy burden to the "Models" module.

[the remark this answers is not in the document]
BLYKIM: I also think that `None` means all models, so it differs from the default model as you mentioned. But `Some(vec![])` means no model, so if possible, I would like to use `Some(DEFAULT_MODEL)`. but if we only compare `None` and `Some(vec![])`, I also prefer `Some(vec![])` with no models.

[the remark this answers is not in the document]
sehkone: The minimum that the migration supports is maintaining the connection between the central server and agents. It's quite certain that we will need to reconfigure agents after the migration. Therefore, `None` is highly preferred, and some meaningless values, such as `u32::MAX`, that can be replaced with default values by agents, are the next preference.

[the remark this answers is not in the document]
BLYKIM: done. However, an error occurred when setting `u64` to `MAX`, so it was set to `u64::MIN` instead.

src/migration/migration_structures.rs
```diff
@@ -509,3 +513,89 @@ impl From<BlockListTlsBeforeV29> for BlockListTlsFields {
...
+    pub log_dir: String,
+    pub dump_dir: String,
+
+    pub dump_items: Option<Vec<DumpItem>>,
+    pub dump_http_content_types: Option<Vec<DumpHttpContentType>>,
+
+    pub giganto_ingest_srv_addr: SocketAddr,
+    pub giganto_name: String,
+
+    pub pcap_max_size: u32, // Unit: MB,
```
sophie-cluml, line 536: Maybe we don't even need to describe the unit, not to reveal info about this field. What do you think?

BLYKIM: good, done

minshao (approved):

src/migration.rs
```diff
@@ -672,13 +662,39 @@ fn migrate_0_29_node(store: &super::Store) -> Result<()> {
...
+                                let port = s.giganto_graphql_port.unwrap_or(u16::MIN);
+                                SocketAddr::new(ip, port)
+                            },
+                        ),
+                        data_dir: String::new(),
+                        log_dir: String::new(),
+                        export_dir: String::new(),
+                        retention: {
+                            let dur_str = format!("{}d", s.retention_period.unwrap_or(u16::MAX));
+                            humantime::parse_duration(&dur_str)?
```
msk, line 691: The current approach of using string formatting and parsing for the `retention` field adds unnecessary complexity. We can simplify this by directly creating a `Duration` object:

```rust
retention: {
    let days = u64::from(s.retention_period.unwrap_or(u16::MAX));
    std::time::Duration::from_secs(days * 24 * 60 * 60)
},
```

This change would make the code more straightforward and easier to understand. It would also remove the need for the `humantime` dependency.

sophie-cluml: I checked if I can push your suggestion to Bly's forked repository, and I could not. Bly will be back from vacation on Thursday, so it looks like he would be able to work on this on Thursday.

msk: Thank you for the update. Given that this is a minor change and to avoid further delay, I'll go ahead and merge this PR. I'll make the suggested modification in a follow-up commit myself.

sophie-cluml: Thank you so much!

PR #309, opened 2024-07-16
Add `language` field in `Account` struct to represent user's language

### LINKED ISSUE #308: Add a new field to `Account` in review-database to represent a user's selected language
To represent a user's selected language, we need to add a new field to Account. 

However, I'm not sure which term would be most appropriate. Is it okay to name this field `language`? If there are no opinion about it, I will proceed with the field name `language`.

div-seungha:
Resolve #308 

It's not related to this PR, but there's one line that I've corrected because I found a typo (`passowrd` -> `password`).
Please review if this is the right way to work on adding `language` field to `Account`. @sehkone @sophie-cluml

### DIFF
--- a/src/account.rs
+++ b/src/account.rs
@@ -35,6 +35,7 @@ pub struct Account {
     pub role: Role,
     pub name: String,
     pub department: String,
+    pub language: Option<String>,
     pub(crate) creation_time: DateTime<Utc>,
     pub(crate) last_signin_time: Option<DateTime<Utc>>,
     pub allow_access_from: Option<Vec<IpAddr>>,
@@ -51,12 +52,14 @@ impl Account {
     /// # Errors
     ///
     /// Returns an error if account creation fails.
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
         username: &str,
         password: &str,
         role: Role,
         name: String,
         department: String,
+        language: Option<String>,
         allow_access_from: Option<Vec<IpAddr>>,
         max_parallel_sessions: Option<u32>,
     ) -> Result<Self> {
@@ -69,6 +72,7 @@ impl Account {
             role,
             name,
             department,
+            language,
             creation_time: now,
             last_signin_time: None,
             allow_access_from,
@@ -283,6 +287,7 @@ mod tests {
             String::new(),
             None,
             None,
+            None,
         );
         assert!(account.is_ok());
 
@@ -298,7 +303,7 @@ mod tests {
     }
 
     #[test]
-    fn account_passowrd_update() {
+    fn account_password_update() {
         let mut account = Account {
             username: "test".to_string(),
             password: SaltedPassword::new_with_hash_algorithm(
@@ -309,6 +314,7 @@ mod tests {
             role: Role::SecurityAdministrator,
             department: String::new(),
             name: String::new(),
+            language: None,
             creation_time: Utc::now(),
             last_signin_time: None,
             allow_access_from: None,
--- a/src/migration.rs
+++ b/src/migration.rs
@@ -38,7 +38,7 @@ use crate::{Agent, AgentStatus, Giganto, Indexed, IterableMap};
 /// // the database format won't be changed in the future alpha or beta versions.
 /// const COMPATIBLE_VERSION: &str = ">=0.5.0-alpha.2,<=0.5.0-alpha.4";
 /// ```
-const COMPATIBLE_VERSION_REQ: &str = ">=0.29.0-alpha.7,<=0.29.0-alpha.7";
+const COMPATIBLE_VERSION_REQ: &str = ">=0.29.0-alpha.8,<=0.29.0-alpha.8";
 
 /// Migrates data exists in `PostgresQL` to Rocksdb if necessary.
 ///
@@ -769,6 +769,7 @@ fn migrate_0_29_account(store: &super::Store) -> Result<()> {
                 role: input.role,
                 name: input.name,
                 department: input.department,
+                language: None,
                 creation_time: input.creation_time,
                 last_signin_time: input.last_signin_time,
                 allow_access_from: input.allow_access_from,
@@ -2225,6 +2226,7 @@ mod tests {
                     role: input.role,
                     name: input.name,
                     department: input.department,
+                    language: None,
                     creation_time: input.creation_time,
                     last_signin_time: input.last_signin_time,
                     allow_access_from: input.allow_access_from,
@@ -2262,6 +2264,7 @@ mod tests {
             Role::SecurityAdministrator,
             "name".to_string(),
             "department".to_string(),
+            None,
             None,
             None,
         )
--- a/src/tables/accounts.rs
+++ b/src/tables/accounts.rs
@@ -169,6 +169,7 @@ mod tests {
             "Department 1".to_string(),
             None,
             None,
+            None,
         )
         .unwrap();
         table.put(&acc1).unwrap();
@@ -180,6 +181,7 @@ mod tests {
             Role::SystemAdministrator,
             "User 2".to_string(),
             "Department 2".to_string(),
+            None,
             None,
             None,
         )
@@ -211,6 +213,7 @@ mod tests {
             "Department 1".to_string(),
             None,
             None,
+            None,
         )
         .unwrap();
         table.put(&acc1).unwrap();
@@ -226,6 +229,7 @@ mod tests {
             Role::SystemAdministrator,
             "User 2".to_string(),
             "Department 2".to_string(),
+            None,
             None,
             None,
         )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   `Node::giganto`.
 - Added `Account::password_last_modified_at` field to track the timestamp of the
   last password modification.
+- Added `Account::language` field to represent user's selected language on the
+  user interface.
 
 ### Changed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "review-database"
-version = "0.29.0-alpha.7"
+version = "0.29.0-alpha.8"
 edition = "2021"
 
 [dependencies]


sehkone:
Could you rewrite your message in English, which is the official language of this organization?

div-seungha:
> Could you rewrite your message in English, which is the official language of this organization?

I just rewrote the message in English.

sehkone:
@div-seungha ,
I think it would be better to ask @sophie-cluml to share how to migrate the previous data. In the meantime, this should be tagged as [WIP].

src/migration.rs
```diff
@@ -2284,6 +2287,7 @@ mod tests {
         let account = res.unwrap();
         if let Some(a) = &account {
             test.password_last_modified_at = a.password_last_modified_at;
+            assert_eq!(test.language, a.language);
```
sophie-cluml, line 2290: To me, it looks like L2292 covers this test. Could you tell me why this line is needed?

div-seungha: I was thinking the same thing as you said and wondering if I should remove that line. But I thought, test codes can act as a test for which function actually works well, but sometimes the test code itself could act as a kind of specification or explanation for engineers to help understanding the function, so I decided that it would be better to reveal the change point of the `Account` field more explicitly. However, since the L2292 already covers this test, I definitely agree this line should be removed when focusing on literal test of the function.

div-seungha: But as I kept thinking about this, I thought it was right to remove the code if it's gone and doesn't affect anything. So I have just pushed the modified version of commit.

div-seungha:
There was a "too many arguments" clippy error in line 54 of `src/account.rs`, so I have just added `#[allow(clippy::too_many_arguments)]` above that line and force-pushed the revision commit. I need to be checked if I modified right. @sophie-cluml

src/migration.rs
```diff
@@ -2262,6 +2264,7 @@ mod tests {
             Role::SecurityAdministrator,
             "name".to_string(),
             "department".to_string(),
+            "".to_string(),
```
sehkone, line 2267: It would be better to assign a default language.

src/tables/accounts.rs
```diff
@@ -167,6 +167,7 @@ mod tests {
             Role::SystemAdministrator,
             "User 1".to_string(),
             "Department 1".to_string(),
+            "language 1".to_string(),
```
sehkone, line 170: Could you use an actual string representing a certain language?

src/tables/accounts.rs
```diff
@@ -180,6 +181,7 @@ mod tests {
             Role::SystemAdministrator,
             "User 2".to_string(),
             "Department 2".to_string(),
+            "language 2".to_string(),
```
sehkone, line 184: The same goes with this.

src/tables/accounts.rs
```diff
@@ -209,6 +211,7 @@ mod tests {
             Role::SystemAdministrator,
             "User 1".to_string(),
             "Department 1".to_string(),
+            "language 1".to_string(),
```
sehkone, line 214: The same goes with this.

src/tables/accounts.rs
```diff
@@ -226,6 +229,7 @@ mod tests {
             Role::SystemAdministrator,
             "User 2".to_string(),
             "Department 2".to_string(),
+            "language 2".to_string(),
```
sehkone, line 232: The same goes with this.

src/migration.rs
div-seungha: Is it correct that your point means setting the default language only within the test code?

sehkone: This is not any test code but the migration code, isn't it?

src/tables/accounts.rs
div-seungha: Is there some specific string form that is agreed for this `language` field? For example, English can simply be referred to as "English", but there can be an abbreviation form such as "en", or an ISO language code form such as "en-US". Or, can I put it in arbitrarily?

minshao: IMHO, It is rather preferred to define a enum for it like the following:
```Rust
#[derive(Default)]
pub enum Language {
    #[default] // sets Korean to be default language
    Korean,
    English,
    ...
  }
```

src/account.rs
```diff
@@ -35,6 +35,7 @@ pub struct Account {
     pub role: Role,
     pub name: String,
     pub department: String,
+    pub language: String,
```
sehkone, line 38: For the initial users who have not set up their languages, it might be better to use `Option`. @sophie-cluml, @div-seungha, could you share your thoughts?

sophie-cluml: That sounds good to me.

div-seungha: I am all for it if we have to consider the initial users.

sehkone: @sophie-cluml @div-seungha 
Thank you for your feedback. Then, the default value would be `None`.

src/tables/accounts.rs
sehkone: Min used one of language standard codes in the UI, so I would like to follow it.

src/migration.rs
div-seungha: I have pushed the revision commit and the type `language` is now defined as `Option<Language>`. I newly defined an enum `language` according to what I was pointed out in the comments below. Also, I found that `Language` was defined in this form of enum in other areas of our work.

src/tables/accounts.rs
div-seungha: On the line of 31~35 in `src/account.rs`, I defined the enum `language`.

src/tables/accounts.rs
div-seungha: As the type of `language` is replaced with `Option<Language>`, That line has been `None`.

src/tables/accounts.rs
div-seungha: As the type of language is replaced with `Option<Language>`, That line has been `None`.

src/tables/accounts.rs
div-seungha: As the type of language is replaced with `Option<Language>`, That line has been `None`.

src/account.rs
div-seungha: I have submitted the revised commit to incorporate all the discussions we have had thus far. I respectfully ask you to review the new code again,  @sehkone @sophie-cluml

div-seungha:
I had submitted a revised commit defined `language` as an enum. But I have just submitted a new commit that changed it back to `Option<string>` in accordance with the opinion that there will be more time consumption when another language is added or changed in the future. @sehkone @sophie-cluml

src/migration.rs
sehkone: @minshao 
@div-seungha was mistaken about my intention that I wanted to use `String` instead of `Language`, so I asked her to get it fixed according to my intention.

src/tables/accounts.rs
sehkone: > @minshao 
> @div-seungha was mistaken about my intention that I wanted to use `String` instead of `Language`, so I asked her to get it fixed according to my intention.

src/tables/accounts.rs
sehkone: > As the type of language is replaced with `Option<Language>`, That line has been `None`.

This is relevant no more.

sehkone (approved):

src/tables/accounts.rs
sehkone: Since the UI uses the combination of ISO 639-1 and ISO 3166-1 codes such as en-US, the UI will send the code combination to REview.

msk:
> I asked people to use `Option<String>`.

While I understand the reasoning for using `String` instead of an `enum` for flexibility, I'm curious about the decision to make it optional, especially considering we already have a default language in our system.

Could you elaborate on why language is implemented as `Option<String>` rather than just `String`? Given that we have a default language, are there specific use cases or scenarios where you anticipate the language might need to be unset or unknown, rather than defaulting to our system's default language?

src/tables/accounts.rs
minshao: I completely agree with you that storing the language as a string will make the database independent of changes to the language and country codes. 

Meanwhile, I think it might be beneficial to store the country code and language code separately in the database. This would allow for easier parsing and manipulation.

msk: Given the complexity and variability of language tags as specified by RFC 4646, I believe a single string representation is the most practical approach. Language tags are not just simple combinations of ISO 639-1 and ISO 3166-1 codes. They can include additional subtags for scripts, regions, and other variants (e.g., "yue-Hant-HK" for Cantonese in traditional script as spoken in Hong Kong, or "gsw-u-sd-chzh" for Swiss German). Parsing and storing these components separately would be cumbersome and error-prone.

minshao: I see. Does that mean parsing the language string to the actual user setting is not our concern? My concern is whether we need to make sure that language string is "legal".

msk: Great question! Given the dynamic nature of language tags, ensuring that a language string is standard-compliant is indeed challenging. The IANA language subtag registry is continuously updated with new tags, making it difficult for us to maintain an exhaustive and up-to-date validation mechanism within the `review-database` crate.

Therefore, keeping the language as a string is likely the best approach for our crate, which is primarily focused on storage. This way, we avoid the complexity and potential pitfalls of trying to validate ever-changing language standards ourselves.

Regarding the legality of the language string, it would be the responsibility of the UI to ensure that only supported and valid language tags are stored in the database. The UI should handle the validation and ensure compliance with the standards it adheres to. Additionally, the UI should define and manage the behavior when encountering language tags retrieved from the database that are not supported by the UI.

While databases typically perform a certain level of validation on the data they receive, in this case, whether a value is valid or not is entirely dependent on the UI. The UI should ensure that only language tags it supports are stored, as the database cannot inherently know which languages the UI has implemented.

sehkone:
> > I asked people to use `Option<String>`.
> 
> While I understand the reasoning for using `String` instead of an `enum` for flexibility, I'm curious about the decision to make it optional, especially considering we already have a default language in our system.
> 
> Could you elaborate on why language is implemented as `Option<String>` rather than just `String`? Given that we have a default language, are there specific use cases or scenarios where you anticipate the language might need to be unset or unknown, rather than defaulting to our system's default language?

@msk

The tasks related to this language have been described in the repository handling repo-unspecific issues as follows.

> ### Tasks
> * [ ]  Add a new field to `Account` in review-database to represent a user's selected language.
> * [ ]  Modify `account`, `insert_account`, `modify_account`, and other related APIs in review-web as needed to support the field.
> * [ ]  Add two new APIs in review-web: one for retrieving the user's language setting, and another for updating it.
> * [ ]  Ensure that the UI retrieves the user's language setting immediately after signing in, stores it in the struct storing information shared between web components, and applies this language setting to both the UI and the user's browser.
> * [ ]  In the UI, implement the menu where the user sets the language preference.
> * [ ]  If no language setting is detected in the browser at the time of user sign-in, default to using the browser’s own language setting.

As put above, the UI will use the browser's language instead of our system's default one if the user hasn't set their language yet. 

> It would be better to assign a default language.

My previous comment, as shown above, might seem to contradict this, but it was intended only for the migration.

sehkone:
> As put above, the UI will use the browser's language instead of our system's default one if the user hasn't set their language yet.

Additionally, the cases where the UI use its language include when the user switches machines or browsers, even if they have already set the language.

msk (approved):

Issue #318, opened 2024-08-05
Add language field into an account update function


div-seungha:
I added a `language` field to the `Account` structure in this commit
https://github.com/petabi/review-database/pull/309

review-web is using the `update` function of review-database to modify contents of accounts. However, the `update` function of review-database did not reflect this change for the `Account` structure. (This is the part I worked on, but I missed it inadvertently.) Therefore, when using the `update` function of review-database in review-web, errors occur when compiling due to insufficient number of arguments.

I created an issue to correct the error. I will upload the PR right away and please review it when the PR comes up.

PR #319, opened 2024-08-05
Add a language update code into the account table update function

div-seungha:
Close: #318 

As I mentioned in the issue, a `language` field was added to the `Account` structure, but the function updating the account did not reflect this. To solve this problem, the following code modification was performed.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new language preference setting for account management, allowing users to specify and manage their language preferences.
	- Enhanced validation ensures consistency when updating language settings for accounts.

- **Chores**
	- Updated versioning to indicate pre-release status (0.29.1-alpha) in the package management file.
	- Enhanced changelog to reflect new features and improve clarity for users and developers.

### DIFF
--- a/src/tables/accounts.rs
+++ b/src/tables/accounts.rs
@@ -76,6 +76,7 @@ impl<'d> Table<'d, Account> {
         role: Option<(Role, Role)>,
         name: &Option<(String, String)>,
         department: &Option<(String, String)>,
+        language: &Option<(Option<String>, Option<String>)>,
         allow_access_from: &Option<(Option<Vec<IpAddr>>, Option<Vec<IpAddr>>)>,
         max_parallel_sessions: &Option<(Option<u32>, Option<u32>)>,
     ) -> Result<(), anyhow::Error> {
@@ -109,6 +110,12 @@ impl<'d> Table<'d, Account> {
                         bail!("old value mismatch");
                     }
                     account.department.clone_from(new);
+                }
+                if let Some((old, new)) = language {
+                    if account.language != *old {
+                        bail!("old value mismatch");
+                    }
+                    account.language.clone_from(new);
                 }
                 if let Some((old, new)) = &allow_access_from {
                     if account.allow_access_from != *old {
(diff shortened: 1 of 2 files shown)


src/tables/accounts.rs
```diff
@@ -76,6 +76,7 @@ impl<'d> Table<'d, Account> {
         role: Option<(Role, Role)>,
         name: &Option<(String, String)>,
         department: &Option<(String, String)>,
+        language: &Option<(Option<String>, Option<String>)>,
```
minshao, line 79: This modification changes public API, thus it should be documented in the CHANGELOG to notify the user. Additionally, please bump alpha version.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "review-database"
-version = "0.29.1"
+version = "0.29.1-alpha"
```
minshao, line 3: Sorry about the confusion, let's leave the version update for later. Could you please revoke the changes on version bump?

div-seungha: I have just revoked the changes related to the version bump, keeping a comment about this PR change in CHANGELOG. Is this the right way to modify it?

src/tables/accounts.rs
div-seungha: I added "unreleased" section in CHANGELOG and wrote a comment about it.

minshao (approved):

Cargo.toml
minshao: Yep~ Thank you

sehkone: Isn't this change breaking? If so, it would be 0.30.0-alpha.0.

sehkone: Oh, you have revoked it. Forget about my comment.

minshao: It is, and you are perfectly right. Considering there's another on-going PR that involves breaking change, I thought it will involve fewer conflicts and make the process a bit simpler, so I dropped the version bump requirement for this specific case.

PR #328, opened 2024-08-19
Add method for get, update in TrustedDomain

### LINKED ISSUE #327: Add method for `update` in TrustedDomain
### Description

Add Rocks Db table's `update` method
- update method to modify with input data

### Reference
https://github.com/aicers/review-web/issues/2

henry0715-dev:
### Desc

Added method `get`, `update` in `trusted_domain`

### Reference

Close #327


sehkone:
I think we need to discuss this further regarding whether `get` is necessary, and others.

henry0715-dev:
> I think we need to discuss this further regarding whether `get` is necessary, and others.

I think it is necessary to satisfy the following conditions.
스크린샷 2024-08-20 오전 8 10 28

sehkone:
> > I think we need to discuss this further regarding whether `get` is necessary, and others.
> 
> I think it is necessary to satisfy the following conditions.

I think the `trusted_domain_list` method of review-web already provides what the conditions require.

src/tables/trusted_domain.rs
```diff
@@ -44,6 +45,31 @@ impl<'d> Table<'d, TrustedDomain> {
...
+    pub fn get(&self, name: &str) -> Result<Option<TrustedDomain>> {
+        let key = name.as_bytes();
+        let Some(value) = self.map.get(key)? else {
+            return Ok(None);
+        };
+
+        Ok(Some(TrustedDomain::from_key_value(key, value.as_ref())?))
+    }
+
+    /// Updates the `trust_domain` in the database.
```
sehkone, line 62: `trust_domain` appears to be a typo. Strictly speaking, `trusted_domain` is not the exact term here because it refers to a module of `tables`. Therefore, either of the following would be proper. In other code, the second is widely used.

```Rust
/// Updates the `trusted_domain` table in the database.

/// Updates the `TrustedDomain` in the database
```

henry0715-dev: @sehkone 
I have completed the change.

src/tables/trusted_domain.rs
```diff
@@ -44,7 +45,18 @@ impl<'d> Table<'d, TrustedDomain> {
...
+    /// Updates the `TrustedDomain` in the database.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the serialization fails or the database operation fails.
+    pub fn update(&self, old: &TrustedDomain, new: &TrustedDomain) -> Result<()> {
+        let (ok, ov) = (old.unique_key(), old.value());
+        let (nk, nv) = (new.unique_key(), new.value());
+        self.map.update((&ok, &ov), (&nk, &nv))
+    }
```
sehkone, line 57: My [comment](https://github.com/petabi/review-database/issues/327#issuecomment-2297985796) is relevant here.

henry0715-dev: The required functions are performed in `self.map.update`.

스크린샷 2024-08-20 오후 2 23 44

sehkone: I thought your questions in [GitHub](https://github.com/petabi/review-database/issues/327#issuecomment-2297713658) and Slack were relevant here, and the two are the same.

The former is my misunderstanding, but why would you ask the latter in Slack even though this `update` suffices?

henry0715-dev: The 'update' function of 'accounts' is also causing an error when changing the name. 
I think it's an issue where old can't be found.
I made an inquiry to determine how to implement the update function in 'trusted Domain' in order to operate normally.

sehkone:
@henry0715-dev 
Could you remove [WIP] that is not necessary any longer? And, I suggest this commit message:

Add the `update` method in `TrustedDomain`

@sophie-cluml 
Could you review this PR?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Extended the `Table<'d, Account>::update` method signature to include the
   `language` parameter, enabling language updates alongside existing fields.
 - Added `category` field to TI db and TI rules.
+- Added `update` method in `trusted_domain`
```
sophie-cluml, line 15:
```suggestion
- Added `update` method in `TrustedDomain`.
```

I suggest above because `TrustedDomain` and `trusted_domain` are different, and I think we need an end mark.

src/tables/trusted_domain.rs
```diff
@@ -79,6 +91,20 @@ mod test {
...
+        let store = setup_store();
+        let table = store.trusted_domain_map();
+        let origin = create_entry("origin");
+        assert!(table.put(&origin).is_ok());
+
+        let updated = TrustedDomain {
+            name: "updated".to_string(),
+            remarks: "updated remarks".to_string(),
+        };
+        assert!(table.update(&origin, &updated).is_ok());
```
sophie-cluml, line 105: I think it would be beneficial to add assertions that check if the `name` and `remarks` are updated correctly after this `update` method is called.

henry0715-dev: I added the test code below.

sophie-cluml: What about adding another assertion that ensures that an entry with the key b"origin" does not exist in the table?

henry0715-dev: I'll also add that information.

sehkone:
> And, I suggest this commit message:
> 
> Add the `update` method in `TrustedDomain`

I think this message should apply to the title of this issue as well.

sehkone (approved):

PR #333, opened 2024-08-28
Added `PasswordPolicy` for managing password expiry period data

### LINKED ISSUE #326: Introduce `PasswordPolicy` for password expiration period
Pre-work is required in "Review-Database" to add password-related logic in "Review-WEB".
The account policy to be added is as follows: b"password expiration period".

**expiry_period_in_secs** This seems to be being used as a value for the expiration of the login session.
I need to save a different value so that you can check the password expiration date.

We need to implement CRUD functionality for the value of "password expiration period." To achieve this, it is necessary to determine whether to add a new field to the AccountPolicy or to declare a separate structure for this purpose.

1. Adding a new field: Modify all the functions that are currently in use.
```
#[derive(Serialize, Deserialize)]
pub struct AccountPolicy {
    pub(crate) expiry_period_in_secs: u32,
    pub(crate) expiry_password_period_in_days: u32,
}
```
2. Adding a new structure
```
#[derive(Serialize, Deserialize)]
pub struct PasswordPolicy {
    pub(crate) expiry_password_period_in_days: u32,
}
```

I would like to ask for your opinion on which approach, between option 1 and option 2, would be preferable for implementation. Henry is scheduled to carry out this task.

reference
https://github.com/aicers/review-web/issues/41
https://github.com/aicers/review-web/issues/42

henry0715-dev:
Close #326

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Added `PasswordPolicy` for managing password expiry period data.
+
 ### Changed
 
 - Extended the `Table<'d, Account>::update` method signature to include the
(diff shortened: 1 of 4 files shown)


sehkone:
Closed because this will be revisited on a solid basis later.

PR #347, opened 2024-09-17
Implement backup functionality for redb

msk:
This change adds the ability to backup the redb database to a file. It also moves all backup-related methods of `Store` to the backup module, in order to keep the lib.rs file clean.


msk:
We will directly use redb.

PR #353, opened 2024-09-27
Add packet attribute triage function

### LINKED ISSUE #354: Implement `score_by_packet_attr` to calculate triage scores properly
The method `score_by_packet_attr` hasn't been implemented yet and needs to be completed.

The score, based on the triage policy, should be evaluated using three criteria:

- whether the confidence is higher than a certain minimum specified in the `TriagePolicy`
- whether the values of each event's attributes fall within the ranges designated in the `TriagePolicy`
- whether certain attribute values of each event match entries in some TI databases

The first criterion is already implemented in the code, while the last one requires further consideration for its design. This issue concerns the second criterion.

kimhanbeom:
@sehkone 
Regarding the addition of packet attribute functionality, I have added this functionality to `DnsCovertChannel` as follows. I would appreciate any feedback on this modification.

closes #354 

## Summary by CodeRabbit

- **New Features**
  - Introduced functions for enhanced attribute comparison, supporting various data types and comparison scenarios.
  - Added dynamic scoring for DNS attributes, improving the accuracy of attribute evaluations.

- **Bug Fixes**
  - Updated method to log warnings for invalid DNS attribute fields, enhancing error handling.

- **Documentation**
  - Clarified method signatures for improved usability and understanding of the API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Added `update_agent_status_by_hostname` public function to `Node` table. This
   function allows updating an agent’s status using the node’s hostname and the
   agent’s ID.
+- Added triage functionality for scoring with attributes of each raw event.
+  Previously, all detection events always returned a score of 0.0. It now
+  compares the values against the provided attribute correctly and returns a
+  score.
 
 ### Changed
 
@@ -22,6 +26,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   instead of Display trait.
 - `Filter::host_name` has been renamed to `Filter::hostname` to align with the
   naming convention used in other parts of the codebase.
+- Modified the `ValueKind` enum to support different types of input for packet
+  attribute triage.
 
 ### Fixed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2024"
 aho-corasick = "1"
 anyhow = "1"
 argon2 = { version = "0.5", features = ["std"] }
+attrievent = { git = "https://github.com/aicers/attrievent.git", tag = "0.2.1" }
 bb8-postgres = { version = "0.9", features = [
   "with-serde_json-1",
   "with-chrono-0_4",
@@ -26,6 +27,7 @@ futures = "0.3"
 humantime-serde = "1"
 ip2location = "0.5.4"
 ipnet = { version = "2", features = ["serde"] }
+memchr = "2"
 num-derive = "0.4"
 num-traits = "0.2"
 postgres-protocol = "0.6"
--- a/src/event.rs
+++ b/src/event.rs
@@ -3435,7 +3435,7 @@ mod tests {
             message: "message".to_string(),
             renewal_time: 100,
             rebinding_time: 200,
-            class_id: vec![4, 5, 6],
+            class_id: "MSFT 5.0".as_bytes().to_vec(),
             client_id_type: 1,
             client_id: vec![7, 8, 9],
             category: EventCategory::InitialAccess,
@@ -3456,7 +3456,7 @@ mod tests {
         let (_, _, syslog_message) = message.unwrap();
         assert_eq!(
             &syslog_message,
-            r#"time="1970-01-01T01:01:01+00:00" event_kind="BlocklistDhcp" category="InitialAccess" sensor="collector1" src_addr="127.0.0.1" src_port="68" dst_addr="127.0.0.2" dst_port="67" proto="17" last_time="100" msg_type="1" ciaddr="127.0.0.5" yiaddr="127.0.0.6" siaddr="127.0.0.7" giaddr="127.0.0.8" subnet_mask="255.255.255.0" router="127.0.0.1" domain_name_server="127.0.0.1" req_ip_addr="127.0.0.100" lease_time="100" server_id="127.0.0.1" param_req_list="1,2,3" message="message" renewal_time="100" rebinding_time="200" class_id="04:05:06" client_id_type="1" client_id="07:08:09""#,
+            r#"time="1970-01-01T01:01:01+00:00" event_kind="BlocklistDhcp" category="InitialAccess" sensor="collector1" src_addr="127.0.0.1" src_port="68" dst_addr="127.0.0.2" dst_port="67" proto="17" last_time="100" msg_type="1" ciaddr="127.0.0.5" yiaddr="127.0.0.6" siaddr="127.0.0.7" giaddr="127.0.0.8" subnet_mask="255.255.255.0" router="127.0.0.1" domain_name_server="127.0.0.1" req_ip_addr="127.0.0.100" lease_time="100" server_id="127.0.0.1" param_req_list="1,2,3" message="message" renewal_time="100" rebinding_time="200" class_id="MSFT 5.0" client_id_type="1" client_id="07:08:09""#,
         );
 
         let blocklist_dhcp = Event::Blocklist(RecordType::Dhcp(BlocklistDhcp::new(
@@ -3467,7 +3467,7 @@ mod tests {
 
         assert_eq!(
             &blocklist_dhcp,
-            r#"time="1970-01-01T01:01:01+00:00" event_kind="BlocklistDhcp" category="InitialAccess" sensor="collector1" src_addr="127.0.0.1" src_port="68" dst_addr="127.0.0.2" dst_port="67" proto="17" last_time="100" msg_type="1" ciaddr="127.0.0.5" yiaddr="127.0.0.6" siaddr="127.0.0.7" giaddr="127.0.0.8" subnet_mask="255.255.255.0" router="127.0.0.1" domain_name_server="127.0.0.1" req_ip_addr="127.0.0.100" lease_time="100" server_id="127.0.0.1" param_req_list="1,2,3" message="message" renewal_time="100" rebinding_time="200" class_id="04:05:06" client_id_type="1" client_id="07:08:09" triage_scores="""#
+            r#"time="1970-01-01T01:01:01+00:00" event_kind="BlocklistDhcp" category="InitialAccess" sensor="collector1" src_addr="127.0.0.1" src_port="68" dst_addr="127.0.0.2" dst_port="67" proto="17" last_time="100" msg_type="1" ciaddr="127.0.0.5" yiaddr="127.0.0.6" siaddr="127.0.0.7" giaddr="127.0.0.8" subnet_mask="255.255.255.0" router="127.0.0.1" domain_name_server="127.0.0.1" req_ip_addr="127.0.0.100" lease_time="100" server_id="127.0.0.1" param_req_list="1,2,3" message="message" renewal_time="100" rebinding_time="200" class_id="MSFT 5.0" client_id_type="1" client_id="07:08:09" triage_scores="""#
         );
     }
 
--- a/src/event/bootp.rs
+++ b/src/event/bootp.rs
@@ -1,10 +1,39 @@
 use std::{fmt, net::IpAddr, num::NonZeroU8};
 
+use attrievent::attribute::{BootpAttr, RawEventAttrKind};
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 
-use super::{EventCategory, LearningMethod, MEDIUM, TriagePolicy, TriageScore, common::Match};
-use crate::event::common::{to_hardware_address, triage_scores_to_string};
+use super::{EventCategory, LearningMethod, MEDIUM, TriageScore, common::Match};
+use crate::event::common::{AttrValue, to_hardware_address, triage_scores_to_string};
+
+macro_rules! find_bootp_attr_by_kind {
+    ($event: expr, $raw_event_attr: expr) => {{
+        if let RawEventAttrKind::Bootp(attr) = $raw_event_attr {
+            let target_value = match attr {
+                BootpAttr::SrcAddr => AttrValue::Addr($event.src_addr),
+                BootpAttr::SrcPort => AttrValue::UInt($event.src_port.into()),
+                BootpAttr::DstAddr => AttrValue::Addr($event.dst_addr),
+                BootpAttr::DstPort => AttrValue::UInt($event.dst_port.into()),
+                BootpAttr::Proto => AttrValue::UInt($event.proto.into()),
+                BootpAttr::Op => AttrValue::UInt($event.op.into()),
+                BootpAttr::Htype => AttrValue::UInt($event.htype.into()),
+                BootpAttr::Hops => AttrValue::UInt($event.hops.into()),
+                BootpAttr::Xid => AttrValue::UInt($event.xid.into()),
+                BootpAttr::CiAddr => AttrValue::Addr($event.ciaddr),
+                BootpAttr::YiAddr => AttrValue::Addr($event.yiaddr),
+                BootpAttr::SiAddr => AttrValue::Addr($event.siaddr),
+                BootpAttr::GiAddr => AttrValue::Addr($event.giaddr),
+                BootpAttr::ChAddr => AttrValue::VecRaw(&$event.chaddr),
+                BootpAttr::SName => AttrValue::String(&$event.sname),
+                BootpAttr::File => AttrValue::String(&$event.file),
+            };
+            Some(target_value)
+        } else {
+            None
+        }
+    }};
+}
 
 #[derive(Serialize, Deserialize)]
 pub struct BlocklistBootpFields {
@@ -182,7 +211,7 @@ impl Match for BlocklistBootp {
         LearningMethod::SemiSupervised
     }
 
-    fn score_by_packet_attr(&self, _triage: &TriagePolicy) -> f64 {
-        0.0
+    fn find_attr_by_kind(&self, raw_event_attr: RawEventAttrKind) -> Option<AttrValue> {
+        find_bootp_attr_by_kind!(self, raw_event_attr)
     }
 }
--- a/src/event/common.rs
+++ b/src/event/common.rs
@@ -5,13 +5,15 @@ use std::{
 };
 
 use anyhow::Result;
+use attrievent::attribute::RawEventAttrKind;
+use bincode::Options;
 use num_traits::ToPrimitive;
 use serde::{Deserialize, Serialize};
 
 use super::{
-    EventCategory, EventFilter, FlowKind, LearningMethod, TrafficDirection, TriagePolicy,
-    eq_ip_country,
+    EventCategory, EventFilter, FlowKind, LearningMethod, TrafficDirection, eq_ip_country,
 };
+use crate::{AttrCmpKind, Confidence, PacketAttr, Ti, ValueKind};
 
 // TODO: Make new Match trait to support Windows Events
 
@@ -30,14 +32,27 @@ pub(super) trait Match {
     fn sensor(&self) -> &str;
     fn confidence(&self) -> Option<f32>;
     fn learning_method(&self) -> LearningMethod;
+    fn find_attr_by_kind(&self, raw_event_attr: RawEventAttrKind) -> Option<AttrValue>;
+    fn score_by_attr(&self, attr_triage: &[PacketAttr]) -> f64 {
+        let total_score = attr_triage.iter().fold(0.0, |score_acc, item| {
+            let Ok(kind) =
+                RawEventAttrKind::from_kind_and_attr_name(&item.raw_event_kind, &item.attr_name)
+            else {
+                return score_acc;
+            };
 
-    /// Calculates a score based on packet attributes according to the triage policy.
-    ///
-    /// Note: This method is currently unused. All implementations return 0.0.
-    /// It's retained for future use as planned by @syncpark.
-    /// For more details, see:
-    /// <https://github.com/petabi/review-database/pull/321#discussion_r1721392271>
-    fn score_by_packet_attr(&self, triage: &TriagePolicy) -> f64;
+            let Some(value) = self.find_attr_by_kind(kind) else {
+                return score_acc;
+            };
+
+            if is_attr_matched(value, item) {
+                score_acc + item.weight.unwrap_or_default()
+            } else {
+                score_acc
+            }
+        });
+        (total_score * 100.0).trunc() / 100.0
+    }
 
     /// Returns whether the event matches the filter and the triage scores. The triage scores are
     /// only returned if the event matches the filter.
@@ -222,9 +237,9 @@ pub(super) trait Match {
                 let triage_scores = triage_policies
                     .iter()
                     .filter_map(|triage| {
-                        let score = self.score_by_ti_db(triage)
-                            + self.score_by_packet_attr(triage)
-                            + self.score_by_confidence(triage);
+                        let score = self.score_by_ti_db(&triage.ti_db)
+                            + self.score_by_attr(&triage.packet_attr)
+                            + self.score_by_confidence(&triage.confidence);
                         if triage.response.iter().any(|r| score >= r.minimum_score) {
                             Some(TriageScore {
                                 policy_id: triage.id,
@@ -245,13 +260,13 @@ pub(super) trait Match {
         Ok((true, None))
     }
 
-    fn score_by_ti_db(&self, _triage: &TriagePolicy) -> f64 {
+    fn score_by_ti_db(&self, _ti_db: &[Ti]) -> f64 {
         // TODO: implement
         0.0
     }
 
-    fn score_by_confidence(&self, triage: &TriagePolicy) -> f64 {
-        triage.confidence.iter().fold(0.0, |score, conf| {
+    fn score_by_confidence(&self, confidence: &[Confidence]) -> f64 {
+        confidence.iter().fold(0.0, |score, conf| {
             if conf.threat_category == self.category()
                 && conf.threat_kind.to_lowercase() == self.kind().to_lowercase()
                 && self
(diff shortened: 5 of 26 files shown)


src/event/dns.rs
```diff
@@ -3,10 +3,33 @@ use std::{fmt, net::IpAddr, num::NonZeroU8};
...
+const DNS_TRANS_ID: &str = "dns-trans_id";
+const DNS_RTT: &str = "dns-rtt";
+const DNS_QCLASS: &str = "dns-qclass";
+const DNS_QTYPE: &str = "dns-qtype";
+const DNS_RCODE: &str = "dns-rcode";
+const DNS_AA_FLAG: &str = "dns-AA";
+const DNS_TC_FLAG: &str = "dns-TC";
+const DNS_RD_FLAG: &str = "dns-RD";
+const DNS_RA_FLAG: &str = "dns-RA";
+const DNS_TTL: &str = "dns-TTLs";
```
sehkone, line 31: I think it would be much better to define the attributes of the enum type. Since the definition is likely to be shared across our modules, we can manage it in a new repository.

kimhanbeom:
@sehkone 
Are there any issues with the implementation other than moving the enum defined for the attributes to a separate repository? If not, I would like to continue with the implementation of packet attribute triage for the remaining detection events.

kimhanbeom:
@sehkone 

Since https://github.com/aicers/attrievent/pull/2 was merged into the main branch of the attrievent repository, I have modified the code to use the enum from that repository.
Now that we are ready to merge this PR into the review-database main branch, could you kindly review the code?

sehkone:
@kimhanbeom, I think it would be better to rename `PacketAttr` to `RawEventAttr`, because other raw events rather than packets are now handled.

src/tables/triage_policy.rs
```diff
@@ -135,6 +138,7 @@ impl Ord for Ti {
 
 #[derive(Clone, PartialEq, Deserialize, Serialize)]
 pub struct PacketAttr {
+    pub raw_event_type: String,
```
sehkone, line 141: How about using its corresponding enum instead of a `String`?

CHANGELOG.md
```diff
@@ -5,6 +5,35 @@ file is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
...
+    from the field in all detection event.
+  - Implemented `score_by_packet_attr` under `Match` trait.
+
+### Changed
+
+- Removed the `tor` module file. The structures (`HttpEventFields`,
+  `TorConnection`) and implementations within that module have been moved to
+  `crate::event::http`.
+- Fixed HTTP detection events to consistently use `referrer` instead of
+  `referrer` and `referrer` interchangeably.
```
sehkone, line 26: I suspect there might be a typo.

src/tables/triage_policy.rs
```diff
@@ -68,8 +68,11 @@ pub enum TiCmpKind {
 #[derive(Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Deserialize, Serialize)]
 pub enum ValueKind {
     String,
-    Integer,
+    SInteger,
+    UInteger,
```
sehkone, line 72: - Could you add some comments to explain the range of these values?
- If `SInteger` means `i32`, don't we need `i64`?
- If `SInteger` is `i32` and `UInteger` is `u32`, I think the pair of `Integer` and `UInteger` would be better.

CHANGELOG.md
```diff
@@ -5,6 +5,35 @@ file is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
...
+- Modified the `ValueKind` enum to support different types of input for packet
+  attribute triage.
+- Changed the type of fields in the detection event structure for some raw
+  event. This change allows users to see meaningful values directly without
+  having to do any special conversion for that field.
+  - `post_body`(HTTP): `Vec<u8>` to `String`.
```
sehkone, line 32: If you mean `HttpThreat::post_body`, I suggest the following:
```
  - `HttpThreat::post_body`: `Vec<u8>` to `String`.
```

src/tables/triage_policy.rs
kimhanbeom: `SInteger` means `i64` and `UInteger` means `u64`. I'll add a comment about that.

CHANGELOG.md
```diff
@@ -5,6 +5,35 @@ file is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
...
 
+## [Unreleased]
+
+### Added
+
+- Added triage functionality for scoring with attributes of each raw event.
+  - Added a new enum type `AttrValue`. This type is used to convert the
+    attribute value of each raw event to its corresponding type to perform
+    comparison operations.
+  - Added the `target_attribute` to the `Match` trait to generate an `AttrValue`
```
sehkone, line 16: I think the `target_attribute` method should be renamed to `to_attr_value`, because it literally returns `AttrValue`.

src/event/dhcp.rs
```diff
@@ -167,9 +212,9 @@ impl BlockListDhcp {
             message: fields.message,
             renewal_time: fields.renewal_time,
             rebinding_time: fields.rebinding_time,
-            class_id: fields.class_id,
+            class_id: String::from_utf8(fields.class_id.clone()).unwrap_or_default(),
```
sehkone, line 215: I think the following is a better choice:

```rust
    class_id: std::str::from_utf8(&fields.class_id).unwrap_or_default().to_string(),
```

src/event/dhcp.rs
```diff
@@ -68,7 +113,7 @@ impl fmt::Display for BlockListDhcpFields {
             self.message.to_string(),
             self.renewal_time.to_string(),
             self.rebinding_time.to_string(),
-            to_hardware_address(&self.class_id),
+            String::from_utf8(self.class_id.clone()).unwrap_or_default(),
```
sehkone, line 116: I suggest:

```rust
    std::str::from_utf8(&fields.class_id).unwrap_or_default().to_string(),
```

src/tables/triage_policy.rs
sehkone: > `SInteger` means `i64` and `UInteger` means `u64`.

In this case, I think the pair of `Integer` and `UInteger` is still better. Besides, `SInteger` even seems to mean "Small Integer".

src/event/dhcp.rs
sehkone: Since `str` has an advantage over `String` in terms of performance, and there will be virtually no cloning if `unwrap` fails, I believe this approach is slightly better.

CHANGELOG.md
kimhanbeom: This part represents the post_body field of all detection events of the HTTP protocol type.

sehkone:
I think I need to explain why this new crate [attrievent](https://github.com/aicers/attrievent) has been introduced.

The kinds of raw events and their attributes change as our software evolves. The purpose of attrievent is to provide a comprehensive list of attributes for both review and the UI simultaneously.

src/event/dhcp.rs
kimhanbeom: done

src/event/dhcp.rs
kimhanbeom: done

CHANGELOG.md
kimhanbeom: done

src/tables/triage_policy.rs
kimhanbeom: done

CHANGELOG.md
kimhanbeom: done

src/tables/triage_policy.rs
kimhanbeom: done

CHANGELOG.md
sehkone:
```
- `post_body` inside structs of `event::http`: `Vec<u8>` to `String`.
```

How about the above for clarity?

msk:
It would be beneficial to break down this PR into smaller, more focused ones. The formatting fixes for old CHANGELOG entries and spelling fixes in the code are great, but they're not directly related to the main changes in this PR.

Would you mind separating those into their own PRs? That way, we can review and merge them independently, making it easier to track changes and test the main functionality of this PR. It'll also make it smaller and more manageable.

sehkone:
> The formatting fixes for old CHANGELOG entries and spelling fixes in the code are great, but they're not directly related to the main changes in this PR.

Regarding the changes to the CHANGELOG, I think I need to share how the team in Seoul recently adopted a method for handling text files such as CHANGELOG and README. The team is now using an extension for VS Code, Prettier, to format these files. As a result, even a small change to either file causes the entire content to be reformatted.

sehkone:
This needs to be rebased, updated to use the release of attrievent, and modified according to the reviewers' comments. Once that is done and [WIP] is removed, the review can proceed.

kimhanbeom:
> This needs to be rebased, updated to use the release of attrievent, and modified according to the reviewers' comments. Once that is done and [WIP] is removed, the review can proceed.

I've fixed it based on your comment above. Could you please review the code?

sehkone:
@kimhanbeom, could you rebase this?

kimhanbeom:
@sehkone 
I have completed the rebase.

src/event.rs
```diff
@@ -3404,7 +3403,7 @@ mod tests {
             message: "message".to_string(),
             renewal_time: 100,
             rebinding_time: 200,
-            class_id: vec![4, 5, 6],
+            class_id: "MSFT 5.0".to_string().into_bytes(),
```
sehkone, line 3406: I think `.as_bytes().to_vec()` would be better if it results in the same outcome.

CHANGELOG.md
```diff
@@ -26,6 +37,18 @@ Versioning](https://semver.org/spec/v2.0.0.html).
...
+  `crate::event::http`.
+- Modified the `ValueKind` enum to support different types of input for packet
+  attribute triage.
+- Changed the type of fields in the detection event structure for some raw
+  event. This change allows users to see meaningful values directly without
+  having to do any special conversion for that field.
+  - `post_body` inside structs of `event::http`: `Vec<u8>` to `String`.
+  - `chaddr` inside structs of `event::bootp`: `Vec<u8>` to `String`.
+  - `class_id`, `client_id` inside structs of `event::dhcp`: `Vec<u8>` to
+    `String`.
```
sehkone, line 51: We can't treat `Vec<u8>` and `String` as the same, because the latter supports UTF-8 characters only.

CHANGELOG.md
```diff
@@ -14,6 +14,17 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Added kind keywords for detailed searching of `RepeatedHttpSessions`,
   `NonBrowser`,`ExternalDdos`, `CryptocurrencyMiningPool`, `BlockListDceRpc`
   detection events.
+- Added triage functionality for scoring with attributes of each raw event.
+  - Introduced a new crate attrievent. The kinds of raw events and their
```
sehkone, line 18: I think `Used the attrievent crate` would be more accurate, because attrievent is an independent crate.

src/event.rs
kimhanbeom: done

CHANGELOG.md
kimhanbeom: done

sehkone:
@kimhanbeom 

I think this needs rebasing. Could you do that for me finishing the review?

kimhanbeom:
@sehkone
I have completed the rebase.

src/event/common.rs
```diff
@@ -30,14 +32,24 @@ pub(super) trait Match {
...
+            };
+            let Some(target_attr_val) = self.to_attr_value(raw_event_attr) else {
+                continue;
+            };
+            if process_attr_compare(target_attr_val, attr) {
+                total_score += attr.weight.unwrap_or_default();
+            }
+        }
+        (total_score * 100.0).trunc() / 100.0
+    }
```
sehkone, line 52: I think this part can be refactored to use the `fold` method for more idiomatic Rust style.

sehkone: (excluding the line 51)

sehkone, line 52: I believe this is one of the most crucial parts to understand the overall logic of this PR. Therefore, I suggest the following naming improvements to help readers better understand how it works.

```rust
    fn score_by_attr(&self, attr_triage: &[PacketAttr]) -> f64 {
        let mut total_score = 0.0;
        for item in attr_triage {
            let Ok(kind) =
                RawEventAttrKind::from_kind_and_attr_name(&item.raw_event_kind, &item.attr_name)
            else {
                continue;
            };
            let Some(value) = self.find_attr_by_kind(kind) else {
                continue;
            };
            if compare_with_attr(value, item) {
                total_score += item.weight.unwrap_or_default();
            }
        }
        (total_score * 100.0).trunc() / 100.0
    }
```

src/event/bootp.rs
```diff
@@ -180,7 +209,7 @@ impl Match for BlockListBootp {
         LearningMethod::SemiSupervised
     }
 
-    fn score_by_packet_attr(&self, _triage: &TriagePolicy) -> f64 {
-        0.0
+    fn to_attr_value(&self, raw_event_attr: RawEventAttrKind) -> Option<AttrValue> {
```
sehkone, line 212: A `to_xxx` method typically implies that it converts the entire `struct` into a certain value, rather than returning a single field. In our case, we are selecting a field that matches a specific type, so I think the following naming would be more intuitive:

```rust
    fn find_attr_by_kind(&self, kind: RawEventAttrKind) ...
```

src/event/bootp.rs
```diff
@@ -1,10 +1,39 @@
...
+use attrievent::attribute::{BootpAttr, RawEventAttrKind};
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 
-use super::{EventCategory, LearningMethod, MEDIUM, TriagePolicy, TriageScore, common::Match};
-use crate::event::common::{to_hardware_address, triage_scores_to_string};
+use super::{EventCategory, LearningMethod, MEDIUM, TriageScore, common::Match};
+use crate::event::common::{AttrValue, to_hardware_address, triage_scores_to_string};
+
+macro_rules! bootp_target_attr {
```
sehkone, line 10: Based on the alternative name `find_attr_by_kind` that I suggested for `to_attr_value`, I think it would be better to rename this accordingly.

```rust
macro_rules! find_attr_by_kind_for_bootp {
```

src/event/common.rs
```diff
@@ -319,15 +331,229 @@ pub fn to_hardware_address(chaddr: &[u8]) -> String {
...
+    VecRaw(&'a [u8]),
+}
+
+fn check_all_or_any<'a, T, F>(attr_val: &'a [T], packet_attr: &'a PacketAttr, compare_fn: F) -> bool
+where
+    F: Fn(&'a T, &'a PacketAttr) -> bool,
+{
+    match packet_attr.cmp_kind {
+        AttrCmpKind::NotEqual | AttrCmpKind::NotContain => {
+            attr_val.iter().all(|x| compare_fn(x, packet_attr))
```
sehkone, line 357: I guess `attr_val.iter().all(|x| !compare_fn(x, packet_attr))` might be what you intended. Could you check it out?

src/event/common.rs
```diff
@@ -319,15 +331,229 @@ pub fn to_hardware_address(chaddr: &[u8]) -> String {
...
+    VecAddr(&'a [IpAddr]),
+    #[allow(dead_code)]
+    VecFloat(Vec<f64>),
+    VecSInt(Vec<i64>),
+    VecUInt(Vec<u64>),
+    VecString(&'a [String]),
+    VecRaw(&'a [u8]),
+}
+
+fn check_all_or_any<'a, T, F>(attr_val: &'a [T], packet_attr: &'a PacketAttr, compare_fn: F) -> bool
```
sehkone, line 351: The current name doesn’t clearly convey what the method actually does. I think the part of "all or any" is not relevant based on the context. Since this checks not just a single attribute value but multiple ones, I believe a more suitable name would be simply `is_matching_list`.

src/event/common.rs
```diff
@@ -319,15 +331,229 @@ pub fn to_hardware_address(chaddr: &[u8]) -> String {
...
+{
+    match packet_attr.cmp_kind {
+        AttrCmpKind::NotEqual | AttrCmpKind::NotContain => {
+            attr_val.iter().all(|x| compare_fn(x, packet_attr))
+        }
+        _ => attr_val.iter().any(|x| compare_fn(x, packet_attr)),
+    }
+}
+
+fn process_attr_compare(target_value: AttrValue, attr: &PacketAttr) -> bool {
```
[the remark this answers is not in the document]
sehkone: I suggest this better one. Sorry for the confusion.

```rust
fn is_attr_matched(value: AttrValue, attr: ...
```

src/event/common.rs
```diff
@@ -319,15 +331,229 @@ pub fn to_hardware_address(chaddr: &[u8]) -> String {
...
+            !((first_val < attr_val) && (second_val >= *attr_val))
+        }
+        (AttrCmpKind::NotRightOpenRange, Some(second_val)) => {
+            !((first_val <= attr_val) && (second_val > *attr_val))
+        }
+        _ => false,
+    }
+}
+
+fn compare_bool_attribute(attr_val: bool, packet_attr: &PacketAttr) -> bool {
```
sehkone, line 467: When a method returns a bool, naming it with `compare_` is not good. I suggest:

```
fn matches_bool_attr(value: bool, packet_attr: &PacketAttr) -> bool {
```

src/event/common.rs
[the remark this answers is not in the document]
sehkone: or

```rust
fn matches_attr(value: AttrValue, attr: &PacketAttr) -> bool {
```

sehkone:
> For the Petabi team, I think I need to explain why this new crate [attrievent](https://github.com/aicers/attrievent) has been introduced.
> 
> The kinds of raw events and their attributes change as our software evolves. The purpose of attrievent is to provide a comprehensive list of attributes for both the REview and the UI simultaneously.

I’d like to add one clarification about the attrievent. Although it provides a list of attributes, it doesn’t guarantee a complete set that covers all of them. This isn’t a significant issue, since we only use the attrievent for triaging events.

src/event/bootp.rs
sehkone: Could you tell me why you employ a macro not a method? I couldn't see any particular advantage.

sehkone: > Could you tell me why you employ a macro not a method? I couldn't see any particular advantage.

I guess the same procedure is applied to different structures, isn't it?

kimhanbeom: Yes, that's precisely the reason

src/event/common.rs
kimhanbeom: For attributes with vector values, depending on the `AttrCmpKind`, it may be necessary to compare and satisfy all values in the vector.
For example, in the case of a `Vec<String>`, if the `AttrCmpKind` is `NotContain`, it should not match any value in the `vec<String>`, so it should be treated as `attr_val.iter().all()`.

src/event/common.rs
```diff
@@ -319,15 +331,229 @@ pub fn to_hardware_address(chaddr: &[u8]) -> String {
...
+        | AttrCmpKind::NotRightOpenRange => {
+            let value = value.as_ref()?;
+            let de_second_value: K = deserialize(value)?;
+            T::try_from(de_second_value).ok()
+        }
+        _ => None,
+    }
+}
+
+fn compare_all_attr_cmp_kind<T>(
```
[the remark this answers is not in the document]
kimhanbeom: I named it `compare_all_attr_cmp_kind` to mean that it can compare all `AttrCmpKinds`, unlike other functions (e.g. `compare_string_attribute`) that have a fixed set of comparable `AttrCmpKind` values.

src/event/common.rs
kimhanbeom: done

src/event/bootp.rs
kimhanbeom: done

src/event/common.rs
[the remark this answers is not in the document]
kimhanbeom: Following your suggestion, I have updated the function name to `is_attr_matched`.

src/event/bootp.rs
kimhanbeom: As per your suggestion, I have updated all macro names to `find_xxx_attr_by_kind`.

src/event/common.rs
kimhanbeom: done

src/event/common.rs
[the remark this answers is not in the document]
kimhanbeom: I have updated the function name to `matches_attr`.

src/event/common.rs
kimhanbeom: As per your suggestion, I have updated all functions that compare attributes and return a boolean type to `matches_xxx_attr`.

sehkone (approved):

sehkone:
@kimhanbeom,

This needs to be rebased to resolve the conflicts.

kimhanbeom:
pending: https://github.com/aicers/attrievent/issues/10

kimhanbeom:
@sehkone
I have completed the rebase.

sehkone:
I believe this PR completes the changes I planned to incorporate.

msk (changes requested):
I have a few observations and suggestions, mostly related to the `CHANGELOG.md` and dependency management, that I think will help with maintainability and our release process.

Regarding the `CHANGELOG.md` changes:

1.  **Formatting Commits:** I noticed quite a few formatting adjustments throughout the `CHANGELOG.md`. Generally, it's good practice to separate purely stylistic or formatting changes into their own commit (or even a separate PR). This helps keep the history clean and makes it easier to review functional changes without the noise of formatting diffs.
2.  **Formatting Rationale & Conventions:** Could you perhaps elaborate on the criteria for the formatting changes? The `review-database` project generally aims to follow `markdownlint`'s default rules, especially for line length (MD013). It wasn't immediately clear if the new formatting aligns with a specific new standard or if it was personal preference.
3.  **Changelog Content and Audience:** The `CHANGELOG.md` for `review-database` is primarily for our users – other developers who consume this crate. We try to focus on user-visible changes like new public APIs, breaking changes, significant feature additions, or bug fixes that impact behavior.
    Some of the added changelog entries, such as detailing the internal mechanics of the `Match` trait (e.g., `find_attr_by_kind`, `score_by_attr` – noting that `Match` is `pub(super)` and not part of the public crate API) or the introduction of internal enums like `AttrValue` (if it's not directly part of the public API for scoring), might be a bit too granular for the main changelog. These details are great for commit messages or the PR description itself.
    On the other hand, changes to public-facing structures like the `ValueKind` enum (which is used in `PacketAttr` for defining triage policies) are definitely important to note, and I see you've included that. We should aim to clearly communicate changes that affect how a user would interact with or configure `review-database`.

Another point is about the dependency on `attrievent`:

1.  **Dependency on Unreleased Version:** The PR currently makes `review-database` depend on a specific commit hash (`rev = "3ffc600"`) of the `attrievent` crate. Our release policy for `review-database` requires that all its dependencies are released versions. This ensures stability and predictability for our users.
2.  **Release Cycle Mismatch:** `attrievent` has a different release cadence than `review-database`. Looking at its history, `attrievent`'s last release was around December 2024, and it hasn't had a new release in the ~4.5 months since. In contrast, `review-database` has had about four releases during that same period. If we merge this PR with `review-database` depending on an unreleased `attrievent` commit, we won't be able to make any new releases of `review-database` until `attrievent` publishes a new version containing these changes. This could significantly block our ability to release updates, bug fixes, or other features for `review-database`.
3.  **Path Forward:** To avoid this, I'd strongly suggest we wait for these `attrievent` changes to be part of an official `attrievent` release. Once that's available, we can update the `Cargo.toml` here to depend on that released version. This would keep our release process smooth and unblocked.

These are mainly process and maintainability points to ensure everything integrates smoothly. The core functionality you've added looks good.

sehkone:
@msk, I'd like to share some context regarding formatting.

The formatting changes in this PR shouldn’t have been included in the first place.  @kimhanbeom should have turned off the Prettier setting that was recently introduced for Seoul engineers.  To avoid confusion, I had already asked @sophie-cluml and her team to disable this setting when working on `review-database`. I assume it may have been overlooked during doing rebases.

Here’s some background on the Prettier setting adopted by the Seoul engineers:

- @sophie-cluml wanted to enforce 80-character line wrapping automatically, and proposed adding `"prettier.proseWrap": "always"` to the VS Code settings.
- This option forces every line to wrap at 80 characters on save.
- The issue is that it causes unnecessary diffs in unchanged parts of the file, confusing PR reviewers.
- For each Aicers repository, we’ve been handling README reformatting separately to avoid noise.
- Specifically for `review-database`, I requested that the setting be turned off whenever developers submit PRs, as I didn’t want this project to be affected by formatting issues.

One of the reasons I didn’t want to apply this broadly was because I wasn’t confident that Prettier strictly follows Markdown conventions.  It doesn’t behave the same way as the VS Code default formatter, and it seems not to fully adhere to all markdown standards—though it clearly handles line wrapping well. Due to time constraints, I haven’t looked deeply into configuring Prettier to follow stricter Markdown rules.

@kimhanbeom, could you revert the README formatting to its previous state by disabling the setting?

sehkone:
@kimhanbeom, regarding the use of `attrievent`’s commit hash:

As you’re well aware, our guidelines—both documented and implied—state that release versions should be used unless there's a specific reason to make an exception. It seems this rule may have been missed during the rebase.

In addition to the Prettier setting not being turned off, this also appears to have been introduced as part of that process.  I’d recommend making it a habit to thoroughly review the code after rebasing to catch these kinds of issues early.

sehkone:
@msk 

I plan to have the Seoul engineers stop using Prettier for Markdown formatting, and will provide them with a detailed guideline on adhering to `markdownlint` rules.

kimhanbeom:
pending: https://github.com/aicers/attrievent/issues/12

kimhanbeom:
@sehkone @msk 
Changes based on the feedback have been applied.

kimhanbeom:
Added migration code for packet attribute.

- Stored attribute info
  image
- Migrating the database to version "0.38.0-alpha.1". (using minireview)
   image
- Migration result
  image

Cargo.toml
```diff
@@ -51,6 +52,7 @@ tokio = { version = "1", features = ["macros"] }
 tokio-postgres-rustls = "0.13"
 toml = "0.8.13"
 tracing = "0.1"
+attrievent = { git = "https://github.com/aicers/attrievent.git", tag = "0.2.0" }
```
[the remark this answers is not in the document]
kimhanbeom: done

sehkone:
I haven’t looked into the migration part yet. I’ll share my thoughts once I’ve completed the full review.

sehkone:
@msk, I've finished reviewing this and I think it looks good to me.

msk:
@kimhanbeom,

I've noticed that this PR introduces several new macros (e.g., `find_bootp_attr_by_kind!`, `find_conn_attr_by_kind!`, etc.) which appear to be used only once in their respective modules, often within functions that seem to act primarily as shallow wrappers for these macro calls.

Could you please share the reasoning behind using macros for these cases? I'm curious to understand if there's a specific benefit or a plan for future use that I might be missing.

Given that the functions calling these macros are essentially just wrappers around the macro call, I'm wondering if inlining the macro's logic directly into those functions might be a more straightforward approach than using a macro in these instances. What are your thoughts on this?

kimhanbeom:
@msk 
The reason for using macros is to handle common processing through macros when different detection event structs share the same fields. (ex) `HttpThreat`, `DomainGenerationAlgorithm`) By implementing it this way, I aimed to reduce redundant code that would otherwise need to be separately implemented for each structure. Even for protocols that currently have only a single detection event defined(ex) `BlocklistBootp`), macros were used to account for possible future additions of events with the same fields.

sehkone:
@msk, I am just letting you know that the last pushed change only relates to updating `attrievent`.

PR #370, opened 2024-11-04
Return an error if the time argument is empty

atyoshioka:
This PR is to return an error if `time` is empty. The `columnStatistics` is defined as follows and `time` must not be empty.

```
columnStatistics(
  cluster: ID!
  time: [NaiveDateTime!]!
  ): JSON!
```

(I'm not sure why the return type is still `JSON`)

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling for empty time inputs in the column statistics retrieval method.
  
- **New Features**
	- Streamlined query construction for fetching column statistics, enhancing performance and clarity.

### DIFF
--- a/src/column_statistics/load.rs
+++ b/src/column_statistics/load.rs
@@ -9,7 +9,7 @@ mod text;
 use std::{cmp::Reverse, collections::HashMap};
 
 use chrono::NaiveDateTime;
-use diesel::{ExpressionMethods, QueryDsl};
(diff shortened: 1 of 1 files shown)


msk:
As far as I understand, `[NaiveDateTime!]!` typically means it can be an empty list (though not null, and each element if present must not be null). Can you help me understand why `time` must not be empty in this case?

atyoshioka:
Thanks! My understanding about GraphQL schema was wrong. Will be closing this.

PR #377, opened 2024-12-09
Rename source to sensor

sophie-cluml:
Closes: #366 

## Summary by CodeRabbit

- **New Features**
	- Updated terminology from "source" to "sensor" across multiple components, improving clarity and consistency.
	- Enhanced migration strategies for database version transitions.

- **Bug Fixes**
	- Addressed inconsistencies in field naming across various data structures and implementations.

- **Documentation**
	- Updated changelog to reflect notable changes and enhancements.

- **Chores**
	- Version updated to `0.33.0-alpha.1` in `Cargo.toml`.

### DIFF
--- a/src/migration.rs
+++ b/src/migration.rs
@@ -38,7 +38,7 @@ use crate::{Agent, AgentStatus, Giganto, Indexed, IterableMap};
 /// // the database format won't be changed in the future alpha or beta versions.
 /// const COMPATIBLE_VERSION: &str = ">=0.5.0-alpha.2,<=0.5.0-alpha.4";
 /// ```
-const COMPATIBLE_VERSION_REQ: &str = ">=0.30.0,<0.33.0-alpha";
+const COMPATIBLE_VERSION_REQ: &str = ">=0.33.0-alpha.1,<0.33.0-alpha.2";
 
 /// Migrates data exists in `PostgresQL` to Rocksdb if necessary.
 ///
@@ -1105,7 +1105,7 @@ fn migrate_outlier_info(store: &super::Store) -> Result<()> {
             timestamp,
             rank,
             id,
-            source,
+            sensor: source,
         };
         let new_key = new_key.to_bytes();
         raw.update((&key, &value), (&new_key, &value))?;
@@ -1758,13 +1758,13 @@ mod tests {
             timestamp,
             rank,
             id,
-            source,
+            sensor: source,
             distance,
             is_saved,
         };
 
         let key = bincode::DefaultOptions::new()
-            .serialize(&(model_id, timestamp, rank, id, &sample.source))
+            .serialize(&(model_id, timestamp, rank, id, &sample.sensor))
             .unwrap();
         let value = bincode::DefaultOptions::new()
             .serialize(&(distance, is_saved))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,18 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Modified `Table::get_patterns` to take a reference to a slice of tuples
   instead of taking ownership of a vector, making the function more flexible.
+- Replaced the term source with sensor, where it refers to a device or software
+  that captures or detects raw events.
+  - This update affects not only the fields in the event protocol structs such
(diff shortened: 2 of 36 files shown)


src/backends/migrations/2024-12-06-125201_rename_event_sources_to_event_sensors/up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE IF EXISTS cluster
+    RENAME COLUMN event_sources TO event_sensors;
```
msk, line 2: Since we're now using 'sensors', which is less ambiguous, would it make sense to simply call the column 'sensors' instead of 'event_sensors'?

src/migration.rs
```diff
@@ -38,7 +38,7 @@ use crate::{Agent, AgentStatus, Giganto, Indexed, IterableMap};
 /// // the database format won't be changed in the future alpha or beta versions.
 /// const COMPATIBLE_VERSION: &str = ">=0.5.0-alpha.2,<=0.5.0-alpha.4";
 /// ```
-const COMPATIBLE_VERSION_REQ: &str = ">=0.30.0,<0.33.0-alpha";
+const COMPATIBLE_VERSION_REQ: &str = ">=0.30.0,<=0.33.0-alpha.1";
```
msk, line 41: Since the new database schema is incompatible with 0.30.0, shouldn't we exclude that version from the compatible range?

src/migration.rs
```diff
@@ -1097,15 +1097,15 @@ fn migrate_outlier_info(store: &super::Store) -> Result<()> {
     let map = store.outlier_map();
     let raw = map.raw();
     for (key, value) in raw.iter_forward()? {
-        let (model_id, timestamp, rank, id, source) = bincode::DefaultOptions::new()
+        let (model_id, timestamp, rank, id, sensor) = bincode::DefaultOptions::new()
```
msk, line 1100: I noticed you've also updated the migration codes to replace 'source' with 'sensor'. However, since 'source' was a valid name in those earlier versions, I think it's unnecessary to rename them in the migration code. If we leave the migration code as is, it would still work correctly and it would also reduce the diff of this PR, making it easier to review.

msk (changes requested):

src/backends/migrations/2024-12-06-125201_rename_event_sources_to_event_sensors/up.sql
sophie-cluml: I think that makes sense. I made the modification.

Should we extend the name simplification to the struct’s field names, `Cluster::event_sources` and `Outlier::event_sources` for consistency? I have not yet worked on the field names, since I think it's safer to ask. Let me know your thoughts!

src/migration.rs
sophie-cluml: It was a significant oversight. Thank you for catching it.

Besides, my current understandings of `COMPATIBLE_VERSION_REQ` are that we update the left-hand side (minimum version) of the range whenever schema changes occur either in RocksDB or PostgreSQL, and that we keep the range as minimal as possible. Based on this, I modified it to `">=0.33.0-alpha.1,<0.33.0-alpha.2"`. If my understanding of the `COMPATIBLE_VERSION_REQ` isn’t fully accurate, I’d greatly appreciate guidance to align better with the intended approach.

src/migration.rs
sophie-cluml: Sounds great! I've ensured the migration-related code remains as is as possible.

src/backends/migrations/2024-12-06-125201_rename_event_sources_to_event_sensors/up.sql
msk: Yes, I think that's a good idea. Please go ahead and update the field names.

src/migration.rs
msk: Right, we should update the minimum version whenever there are schema changes.

src/backends/migrations/2024-12-06-125201_rename_event_sources_to_event_sensors/up.sql
sophie-cluml: Thank you, done.

msk (approved):

Issue #384, opened 2024-12-19
Fix migration failure in 0.33.0


sophie-cluml:
In #377, the PostgreSQL schema was updated while the RocksDB schema remained unchanged. However, the `migrate_data_dir` function was not updated to adjust the VERSION file accordingly. This led to migration failure in version 0.33.0.

I think we would need to yank 0.33.0.

PR #385, opened 2024-12-19
Fix migration error

sophie-cluml:
Close: #384 

-------

I modified the `version` in the Cargo.toml and `COMPATIBLE_VERSION_REQ` and `migrate_data_dir` in migration.rs, under the assumption that 0.33.0 will be yanked, and that 0.33.1 will be the next release, with this quick fix.

I did not mark the 0.33.0 as YANKED in this PR, because from my understanding so far, that should not be handled in this PR. But if I should mark it as YANKED in the CHANGELOG in this PR, please let me know.

### DIFF
--- a/src/migration.rs
+++ b/src/migration.rs
@@ -38,7 +38,7 @@ use crate::{Agent, AgentStatus, Giganto, Indexed, IterableMap};
 /// // the database format won't be changed in the future alpha or beta versions.
 /// const COMPATIBLE_VERSION: &str = ">=0.5.0-alpha.2,<=0.5.0-alpha.4";
 /// ```
-const COMPATIBLE_VERSION_REQ: &str = ">=0.33.0,<0.34.0-alpha";
+const COMPATIBLE_VERSION_REQ: &str = ">=0.30.0,<0.34.0-alpha";
 
 /// Migrates data exists in `PostgresQL` to Rocksdb if necessary.
 ///
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ file is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
 this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Corrected the migration failure issue. The previous version, 0.33.0, needs to
+  be yanked due to this issue.
+
 ## [0.33.0] - 2024-12-12
 
 ### Changed
@@ -740,6 +747,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Modified `FtpBruteForce` by adding an `is_internal` field which is a boolean
   indicating whether it is internal or not.
 
+[Unreleased]: https://github.com/petabi/review-database/compare/0.33.0...main
 [0.33.0]: https://github.com/petabi/review-database/compare/0.32.0...0.33.0
 [0.32.0]: https://github.com/petabi/review-database/compare/0.31.0...0.32.0
 [0.31.0]: https://github.com/petabi/review-database/compare/0.30.0...0.31.0
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "review-database"
-version = "0.33.0"
+version = "0.33.1-alpha.1"
 edition = "2021"
 
 [dependencies]


minshao:
> Close: #384
> 
> I modified the `version` in the Cargo.toml and `COMPATIBLE_VERSION_REQ` and `migrate_data_dir` in migration.rs, under the assumption that 0.33.0 will be yanked, and that 0.33.1 will be the next release, with this quick fix.
> 
> I did not mark the 0.33.0 as YANKED in this PR, because from my understanding so far, that should not be handled in this PR. But if I should mark it as YANKED in the CHANGELOG in this PR, please let me know.

---

### Suggested Changes to Address Compatibility Issue

Thank you for your efforts in addressing this issue! While using a dummy function in this case does resolve the problem, it is worth noting that, as per our convention, we typically handle such cases by silently migrating to version `0.33` for RocksDB.

#### Proposed Adjustment:
In the `0.33.0` release, the current implementation sets the `COMPATIBLE_VERSION_REQ` constant as follows:
```rust
const COMPATIBLE_VERSION_REQ: &str = ">=0.33.0, <0.34.0-alpha";
```

However, for consistency and to ensure broader compatibility, it should instead be defined as:
```rust
const COMPATIBLE_VERSION_REQ: &str = ">=0.30.0, <0.34.0-alpha";
```

#### Reason for the Adjustment:
The rationale for using `>=0.30.0` is that version `0.30.0` was the last version where the RocksDB format was modified. This ensures backward compatibility with older versions while still adhering to our standard convention.

By adopting this approach, we maintain consistency across versions and avoid introducing unnecessary discrepancies in our version compatibility management.

---

Let me know if further clarification is needed! 😊

sehkone:
Since the PostgreSQL schema has changed even though RocksDB hasn't, shouldn't the compatible version be updated accordingly? Or does the compatible version depend solely on RocksDB?

minshao:
> Since the PostgreSQL schema has changed even though RocksDB hasn't, shouldn't the compatible version be updated accordingly? Or does the compatible version depend solely on RocksDB?

* The compatible version depends solely on RocksDB. 
* `migrate_backend` here is used for migration that needs data transfer from PosgreSQL to RocksDB. And it will kept here until we migrate to RocksDB completely.
* PostgreSQL migration is handled in *src/backends/migration*

sehkone:
> > Since the PostgreSQL schema has changed even though RocksDB hasn't, shouldn't the compatible version be updated accordingly? Or does the compatible version depend solely on RocksDB?
> 
> * The compatible version depends solely on RocksDB.
> * `migrate_backend` here is used for migration that needs data transfer from PosgreSQL to RocksDB. And it will kept here until we migrate to RocksDB completely.
> * PostgreSQL migration is handled in _src/backends/migration_

I appreciate your explanation. In fact, I didn’t fully understand the plan to completely migrate to RocksDB and stop using PostgreSQL. Thank you again for letting me know about it.

sophie-cluml:
Thank you very much for kind explanation. Below are done:

- modified `COMPATIBLE_VERSION_REQ` to ">=0.30.0,<0.34.0-alpha"
- removed `migrate_0_30_to_0_33_1`

src/migration/migration_structures.rs
```diff
@@ -1882,3 +1882,27 @@ pub struct GigantoConfig {
 
     pub ack_transmission: u16,
 }
+
+#[derive(Debug, PartialEq)]
+pub struct OutlierInfoKeyBefore33 {
```
minshao, line 1887: Could you clarify the reasoning behind adding the structure definition? Based on my understanding, the change from `OutlierInfo::source` to `OutlierInfo::sensor` is purely a variable name update and does not require any data migration since it doesn’t impact database-related data. As such, it seems like data migration might not be necessary in this case. Could you provide some insight into this decision?

sophie-cluml: Following advice from @sehkone, I introduced a new struct to preserve the original naming, as this might be beneficial for some users. The rationale behind this was that the fact that the variable name change does not impact the RocksDB schema might not be obvious to some users. It is my understanding that it was Kone's main reasoning for this advice, so I would like to invite him if it needs to be clarified further. Besides, Sehkone and I are aware that this is purely a name change and does not alter the RocksDB schema or the data layout, so this choice is not based on the idea that it accompanies schema change. 

If this struct definition causes more confusion than clarity, I am happy to remove it—please let me know! I truly appreciate your time and feedback.

sehkone: Because I asked @sophie-cluml to handle this, I am answering it now. This does not entail any migration, for the exact reason you mentioned. This struct is required only because the current code has recently been changed and needs the previous definition to avoid compilation errors unless the previous migration code is also modified. The last code resolved this by modifying the prior migration, which I believe could mislead people into thinking changes occurred when they actually did not.

sehkone: > If this struct definition causes more confusion than clarity, I am happy to remove it—please let me know! I truly appreciate your time and feedback.

So am I!

src/migration.rs
```diff
@@ -38,7 +38,7 @@ use crate::{Agent, AgentStatus, Giganto, Indexed, IterableMap};
 /// // the database format won't be changed in the future alpha or beta versions.
 /// const COMPATIBLE_VERSION: &str = ">=0.5.0-alpha.2,<=0.5.0-alpha.4";
 /// ```
-const COMPATIBLE_VERSION_REQ: &str = ">=0.33.0,<0.34.0-alpha";
+const COMPATIBLE_VERSION_REQ: &str = ">=0.30.0,<0.34.0-alpha";
```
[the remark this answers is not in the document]
sophie-cluml: Aha, thank you for the guide. I modified it to `">=0.30.0,<=0.33.1-alpha.1"`.

I have one question though. Setting the `COMPATIBLE_VERSION_REQ` as above causes `migration::tests::version` test to fail. I have thought about a few ways to address this issue, but I think it is better to seek your guidance. So I just left the test to fail. https://github.com/petabi/review-database/actions/runs/12426679664/job/34695318406?pr=385 Could you share your insight in this aspect please?

src/migration/migration_structures.rs
minshao: Thank you both for the detailed explanations—I truly appreciate your insights!

Regarding the variable name change, I noticed that it is already well-documented in the CHANGELOG. To keep the codebase as clean and minimal as possible, I suggest removing this change from the code itself, as the documentation seems sufficient.

Additionally, I have a concern about the structure name `OutlierInfoKeyBefore33`. While I understand its intent—to avoid database manipulation and note the name change timestamp—the name could potentially cause confusion. Specifically, it is used in `migrate_outlier_info`, which is primarily part of the `migrate_0_26_to_0_28` logic. A name like `OutlierInfoKey28` might be more intuitive in this context, as the current naming implies a future reference (`Before33`) for something tied to an earlier migration. 

That said, if the team feels it’s important to document this timestamp explicitly in the code, I’m okay with proceeding with the current `OutlierInfoKeyBefore33` as well. Let me know what you think!

sophie-cluml: Thank you for sharing your thoughts, and I removed it.

src/migration.rs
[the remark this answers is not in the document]
minshao: Apologies for the confusion earlier! Upon reviewing, I realize I was mistaken. Your previous commit was correct. The proper values should be:

- In **Cargo.toml**: `version = "0.33.1-alpha.1"`  
- In **src/migration.rs**: `const COMPATIBLE_VERSION_REQ: &str = ">=0.30.0,<0.34.0-alpha";`

Thanks for your patience!

minshao (approved):

Issue #399, opened 2025-01-14
Modify the node structure to support Mitre infomation module


kimhanbeom:
As the Mitre information module has been modified to support remote configuration, the review-database must be modified to manage that module as a node.
- Modify node struct and its associated code.
- Add migration function for changed node struct.

kimhanbeom:
We decided to add a separate structure, `Remote`, to store configuration for all of our remote servers. (https://github.com/petabi/review-database/pull/401#discussion_r1951950143)

Here is the direction of the modification.
- Add a `Remote` structure and storing its information in the db
- Add a `remotes` field of type `Vec<Remote>` inside the `Node` structure (the `giganto` field is removed and the information from this field is also stored in `remotes`).
- Add migration function for changed node struct.

PR #401, opened 2025-01-23
Modify the `Node` structure

kimhanbeom:
- Modify the `Node` structure to allow you to manage the configuration of all remote servers that communicate with the REview.
   - Introduces `Remote`, a new structure that stores configuration information for remote servers. The data in `Remote` is stored via `Table<Remote>`.
   - Add a `remotes` field of type `Vec<Remote>` within the `Node` structure to store the all remote server configuration.
   - Add a `Datalake`, `TiContainer` field inside `node::kind`.
   - Modify migration function `migrate_0_29_node`.
   - Add migration function`migrate_0_35_node`.

Close: #399

## Migration Test
- Insert a node with version "0.27.1".
  image
- Migrating the database to version "0.35.0-alpha.1". (using minireview)
  image
- Migration result.
  - `nodeList` query 
     image
  - query result
    image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,18 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   naming convention used in other parts of the codebase.
 - Modified the `ValueKind` enum to support different types of input for packet
   attribute triage.
+- Modified the `Node` structure to support managing the configuration of all
+  external services that communicate with REview.
+  - Introduced `ExternalService`, a new structure that stores configuration
+    drafts for external services.
+  - Introduced `ExternalServiceConfig`, `ExternalServiceStatus`, and
+    `ExternalServiceKind` as types used by the `ExternalService` structure.
+    - `ExternalServiceConfig` and `ExternalServiceStatus` are aliases of
+      existing types (`node::Config`, `node::Status`).
+    - `ExternalServiceKind` is a newly introduced enum with variants such as
+      `DataStore` and `TiContainer`.
+  - Added an `external_services` field of type `Vec<ExternalService>` to the
+    `Node` structure to store all external service configurations.
 
 ### Fixed
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -66,10 +66,11 @@ pub use self::tables::{
     AccessToken, AccountPolicy, Agent, AgentConfig, AgentKind, AgentStatus, AllowNetwork,
     AllowNetworkUpdate, AttrCmpKind, BlockNetwork, BlockNetworkUpdate, Confidence,
     CsvColumnExtra as CsvColumnExtraConfig, Customer, CustomerNetwork, CustomerUpdate, DataSource,
-    DataSourceUpdate, DataType, Filter, Giganto, IndexedTable, Iterable, ModelIndicator, Network,
-    NetworkUpdate, Node, NodeProfile, NodeTable, NodeUpdate, OutlierInfo, OutlierInfoKey,
-    OutlierInfoValue, PacketAttr, ProtocolPorts, Response, ResponseKind, SamplingInterval,
-    SamplingKind, SamplingPeriod, SamplingPolicy, SamplingPolicyUpdate, Structured,
+    DataSourceUpdate, DataType, ExternalService, ExternalServiceConfig, ExternalServiceKind,
+    ExternalServiceStatus, Filter, IndexedTable, Iterable, ModelIndicator, Network, NetworkUpdate,
+    Node, NodeProfile, NodeTable, NodeUpdate, OutlierInfo, OutlierInfoKey, OutlierInfoValue,
+    PacketAttr, ProtocolPorts, Response, ResponseKind, SamplingInterval, SamplingKind,
+    SamplingPeriod, SamplingPolicy, SamplingPolicyUpdate, Structured,
     StructuredClusteringAlgorithm, Table, Template, Ti, TiCmpKind, Tidb, TidbKind, TidbRule,
     TorExitNode, TrafficFilter, TriagePolicy, TriagePolicyUpdate, TriageResponse,
     TriageResponseUpdate, TrustedDomain, TrustedUserAgent, UniqueKey, Unstructured,
@@ -260,6 +261,12 @@ impl Store {
     #[allow(clippy::missing_panics_doc)]
     pub fn outlier_map(&self) -> Table<OutlierInfo> {
         self.states.outlier_infos()
+    }
+
+    #[must_use]
+    #[allow(clippy::missing_panics_doc)]
+    pub fn external_service_map(&self) -> Table<ExternalService> {
+        self.states.external_service()
     }
 
     #[must_use]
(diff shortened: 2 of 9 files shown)


src/tables/node.rs
```diff
@@ -14,7 +14,7 @@ use crate::{
 };
 
 #[derive(Default, Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub struct Giganto {
+pub struct ConnectionlessAgent {
```
sehkone, line 17: While `Agent` has a key for each instance, `ConnectionlessAgent` does not. Additionally, `Agent`s are represented as a `Vec<Agent>`, whereas `ConnectionlessAgent`s are instantiated individually, like `giganto` and `ti_container`. Furthermore, `ConnectionlessAgent` includes its status through the `AgentStatus` field. These differences highlight inconsistencies, and on top of that, the name `ConnectionlessAgent` doesn't seem ideal to convey the true meaning. I believe we need a better approach to address these inconsistencies and clarify the concept.

sehkone: @sophie-cluml @kimhanbeom 

As we discussed, it would be nice to introduce the `Remote` struct to handle remote servers that do not use QUIC communication. It will be similar to `Agent` in many ways.

kimhanbeom:
@sehkone 
I have completed the code modification to handle remote server configuration via the 'Remote' structure.

sehkone:
> @sehkone I have completed the code modification to handle remote server configuration via the `Remote` structure.

Thank you for what you've done. I will review the code soon.

sehkone:
@msk,

The delay is due to this PR requiring immediate changes to the UI. I’ll check the UI-side schedule and revisit this very soon.

sehkone:
@kimhanbeom,

Since the UI will support the required features as we discussed, this PR should be finalized very soon. Could you rebase it and check if any further updates are needed?

kimhanbeom:
Rebase completed.
- Migrating the database to version "0.38.0-alpha.1". (using minireview)
  image
- Migration result.
  image

kimhanbeom:
This PR should not merged until this issue(https://github.com/petabi/review-database/issues/439) is completed for the product alpha.6 release.

sehkone:
I think I’ll need a couple of days to review this PR thoroughly. Once I’ve completed the review, I’ll request approval for the merge.

sehkone:
@sophie-cluml, could you review this?

src/tables/node.rs
```diff
@@ -332,14 +486,21 @@ impl Iterator for TableIter<'_> {
                     }
                 }
 
+                let mut remotes = vec![];
+                for aid in inner.remotes {
```
sophie-cluml, line 490: I think the code in L483 used `aid` for agent id, so maybe `rid` might be relevant here for remote id?

CHANGELOG.md
```diff
@@ -20,6 +20,13 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   a breaking change if you were using identifiers affected by this change.
 - Modify the syslog message generation procedure to use dedicated formatter
   instead of Display trait.
+- Modified the `Node` structure to allow you to manage the configuration of all
+  remote servers that communicate with the REview.
+  - Introduced `Remote`, a new structure that stores configuration information
```
sophie-cluml, line 25: Just a minor suggestion: I think the reason we used `Giganto` or introduced `Remote` (as distinct from `Agent`) is that `Remote` is meant to store only configuration drafts, so it might be helpful to make that intention explicit — perhaps by mentioning that `Remote` "stores configuration drafts only". Of course, for the sake of simplicity, I think it’s also good to leave it as it is. Just sharing my thought!

sophie-cluml (approved):

sehkone:
@msk,

At my request, @kimhanbeom has built this on top of #353.

src/lib.rs
```diff
@@ -63,12 +63,12 @@ pub use self::migration::{migrate_backend, migrate_data_dir};
 pub use self::model::{Digest as ModelDigest, Model};
 use self::tables::StateDb;
 pub use self::tables::{
-    AccessToken, AccountPolicy, Agent, AgentConfig, AgentKind, AgentStatus, AllowNetwork,
-    AllowNetworkUpdate, AttrCmpKind, BlockNetwork, BlockNetworkUpdate, Confidence,
+    AccessToken, AccountPolicy, Agent, AgentRemoteConfig, AgentRemoteKind, AgentRemoteStatus,
```
msk, line 66: Could you share some context on why these changes were necessary? Since these are part of the public API, they will be breaking changes for existing users of `review-database`.

sehkone: To avoid affecting the existing structure or introducing breaking changes, are you suggesting adding new types — `RemoteConfig`, `RemoteKind`, and `RemoteStatus`—while keeping the existing `AgentConfig`, `AgentKind`, and `AgentStatus`?

If so, I'm on the same page—I should've looked more closely into this part.

msk: My apologies if my initial question wasn't perfectly clear. My main intention was to understand the reasoning and context behind renaming `AgentConfig` to `AgentRemoteConfig` (and the related `Kind` and `Status` types). Since these are part of the public API and would be breaking changes for existing users of `review-database`, I wanted to make sure I understood the problem these changes are solving before discussing specific solutions.

Based on the changes, my current interpretation is that `AgentRemoteConfig` might be intended to handle configurations for both what we currently call 'agents' and also another type of entity, which these new names are perhaps designating as 'remote'. Is that understanding roughly correct?

If so, I'm a little concerned that the prefix `AgentRemote` could be a source of confusion. To me, it reads more like 'an agent that is remote' or 'a remote that is also an agent,' rather than perhaps 'a configuration that could apply to an agent OR a remote entity.' If the goal is to represent something that can be *either* an agent *or* this other 'remote' concept, a different naming strategy might be clearer. For instance, we might consider a new encompassing term or a more explicit naming convention.

Similarly, I'm trying to get a clear picture of what 'remote' distinguishes in this new terminology. My guess from the context is that it might be used for non-agent daemons or services. If that's the idea, I fear 'remote' could be ambiguous because many of our existing agents are already 'remote' in the common meaning of the word (i.e., they operate on different machines from the `review-database` itself). This could make it tricky for developers to intuitively grasp what 'remote' specifically refers to in `AgentRemoteStatus` versus just a standard agent's status that happens to be reported from a remote machine.

My main thought here is that we want to ensure the API's terminology is as intuitive and unambiguous as possible for all developers who will be using it. Clear naming can save a lot of headaches and misunderstandings down the road! And, of course, it's entirely possible I'm missing some important background or previous discussions that led to these choices, which is why I'm keen to understand the perspective from your side.

Perhaps you could elaborate a bit on the specific distinction you're aiming to capture with `AgentRemote` and `Remote`?

kimhanbeom:
@msk @sehkone 
- Removed the intermediate commit related to #353.
- To avoid introducing breaking changes to the public API while making their use in the remote context clear, the existing `Status`, `Config`, and `Kind` types were aliased as `RemoteStatus`, `RemoteConfig`, and `RemoteKind`.

sehkone:
@kimhanbeom,

Could you do what I mentioned in [this comment](https://github.com/petabi/review-database/pull/447#issuecomment-2954039098), including rebasing?

Issue #322, opened 2024-08-12
Fix learning method matches for the detected events


syncpark:
In **src/event.rs**, the learning methods are defined as follows:
```rust
/// in src/event.rs
/// Machine Learning Method.
#[derive(Clone, Copy, Eq, PartialEq, Deserialize, Serialize)]
pub enum LearningMethod {
    Unsupervised,
    SemiSupervised,
}
```

After the `EventCategory::HttpThreat` is replaced to `EventCategory::Reconnaissance`, the following code in **src/event/common.rs** may return incorrect results.
https://github.com/petabi/review-database/blob/2eda8f9ea7ddc8027df6a7ed5991af7db4a80990/src/event/common.rs#L163-L172

## Tasks:
- The matching code must be modified to return correct results.
- And the learning methods of the detected events need to be assigned when the detected events are generated.

syncpark:
To solve this issue, all event structures must contain LearningMethod value and implement `learning_method()` function.

PR #406, opened 2025-02-26
Fix learning method matches for the detect events

kimhanbeom:
The previous implementation of learning method match used the `EventCategory`, but now all detected events have been modified to return `LearningMethod` (via the new `learning_method` method added to `Match` trait), so learning method match is performed using this value.

Close: #322


kimhanbeom:
@sehkone @syncpark 
 Could you please review the this pr?

syncpark (approved):

Issue #408, opened 2025-03-07
Add a field for customer id in `Account`


kimhanbeom:
## Background
- We have a requirement that based on the login account, only data belonging to that account (raw events, detection events, etc.) should be visible in the UI.
- Accordingly, the account information structure managed by review-database needs to be modified to include customer information, and the modification direction is as follows.

## Task
- Add a `customer_ids` field of type `Option<Vec<u32>>` to store customer information in the `Account` structure.
   - `Account` can have one or more customer id.
   - If `Account` does not have any customer id(None), it means that it has all customer id's  (This is only true for the Admin account (`SystemAdministrator`)).
- Modify `Account` insert/update code.
- Add `Account` migration code.

PR #409, opened 2025-03-10
Add a field for customer id in Account

kimhanbeom:
- migration
  스크린샷 2025-03-10 오전 10 51 35

- migration result 
스크린샷 2025-03-10 오전 10 52 19

Close: #408

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Changed
+
+- Added `Account::customer_ids` field to store the customer id that the user
+  belongs to.
+
 ## Fixed
 
 - Fixed learning method matches for the detected events. The previous
--- a/src/migration.rs
+++ b/src/migration.rs
@@ -99,7 +99,7 @@ use crate::{Agent, AgentStatus, Giganto, Indexed, IterableMap};
 /// // release that involves database format change) to 3.5.0, including
 /// // all alpha changes finalized in 3.5.0.
 /// ```
-const COMPATIBLE_VERSION_REQ: &str = ">=0.34.0,<0.36.0";
+const COMPATIBLE_VERSION_REQ: &str = ">=0.36.0-alpha.1,<0.36.0-alpha.2";
 
 /// Migrates data exists in `PostgresQL` to Rocksdb if necessary.
 ///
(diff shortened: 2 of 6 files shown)


kimhanbeom:
@sehkone 
Could you please review the current pull request?

sophie-cluml:
I think this PR suits our need.

src/migration.rs
```diff
@@ -277,6 +282,26 @@ fn read_version_file(path: &Path) -> Result<Version> {
     Version::parse(&ver).context("cannot parse VERSION")
 }
 
+fn migrate_0_34_0_to_0_36_alpha_1(store: &super::Store) -> Result<()> {
```
sehkone, line 285: This should be `migrate_0_34_0_to_0_36` according to [the guideline](https://github.com/petabi/review-database/blob/2870c28c48ca4094415e90c89647686a5473a8b2/src/migration.rs#L72-L90).

src/migration.rs
```diff
@@ -277,6 +282,26 @@ fn read_version_file(path: &Path) -> Result<Version> {
...
+    migrate_0_36_alpha_1_account(store)
+}
+
+fn migrate_0_36_alpha_1_account(store: &super::Store) -> Result<()> {
```
sehkone, line 289: Accordingly, this should be `migrate_0_36_account` without `alpha_1`.

src/migration.rs
```diff
@@ -4109,4 +4148,48 @@ mod tests {
         let settings = TestSchema::new_with_dir(db_dir, backup_dir);
         assert!(super::migrate_0_30_to_0_34_0(&settings.store).is_ok());
     }
+
+    #[test]
+    fn migrate_0_36_alpha_1_account() {
```
sehkone, line 4153: `alpha_1` needs to be removed.

src/migration.rs
kimhanbeom: done

src/migration.rs
kimhanbeom: done

src/migration.rs
kimhanbeom: done

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Changed
+
+- Added `Account::customer_ids` field to store the customer id the user belongs
```
sehkone, line 12: It would be better to add 'that' between 'customer id' and 'the user' for clarity.

kimhanbeom: done

src/migration/migration_structures.rs
```diff
@@ -2102,3 +2104,64 @@ pub struct GigantoConfig {
 
     pub ack_transmission: u16,
 }
+
+#[derive(Clone, Deserialize, Serialize, PartialEq, Debug)]
+pub struct AccountBeforeV35 {
```
sehkone, line 2109: @kimhanbeom @sophie-cluml,

Since the new `Account` is introduced in 0.36.0, I think the old one should be named `AccountBefore36`. I wonder if the current code follows this way.

kimhanbeom: I have renamed `AccountBeforeV35` to `AccountBeforeV36`.

sehkone (approved):

Issue #411, opened 2025-03-18
Modify `dest_addr`, `src_addr` of Match trait to `Vec<IpAddr>` for detection event detail search


kimhanbeom:
Related issue: QA repository issue 82

- When viewing the detection event details in the UI, the event item does not display the information of the detection event.
   <img width="1265" alt="Image" src="https://github.com/user-attachments/assets/8dff073d-8fd2-40ba-b68b-425b1ce96411" />
- The cause of this issue is as follows.
  - To get a detailed view of a specific detection event in the UI, we need to set the src addr(source) and dst addr(destination) of the detection event as filter values and call the `eventList` API again.
  - If the detection event has multiple addresses (src or dst), we are only setting one address in the filter value.
  - review-database handles filtering detection events by comparing `source`(src_addr), `destination`(dst_addr) in filter values with `src_addr`, `dst_addr` methods in `Match` trait
  - But for certain detection events with multiple addresses, we currently use `IpAddr::V4(Ipv4Addr::UNSPECIFIED))` as the return value of `src_addr` and `dst_addr`.
    ```rust
       fn src_addr(&self) -> IpAddr {
          IpAddr::V4(Ipv4Addr::UNSPECIFIED)
       }
       
       fn dst_addr(&self) -> IpAddr {
          IpAddr::V4(Ipv4Addr::UNSPECIFIED)
       } 
    ```
- To fix this, the following modifications are required.
  - Modify the return type of the `src_addr`, `dst_addr` methods in the `Match` trait to `Vec<IpAddr>>`.
  - Modify the return values of `src_addr`, `dst_addr` methods for detection events with multiple addresses to use the src addr and dst addr of the detection event instead of `IpAddr::V4(Ipv4Addr::UNSPECIFIED))`.
  - As the return types of the `src_addr` and `dst_addr` methods change, Modify the code in the part that filters detection events accordingly.
- The following detection events are not currently shown in the event details.
  - `ExternalDdos`
  - `MultiHostPortScan`
  - `RdpBruteForce`

PR #413, opened 2025-03-24
Fix the ability to filtering event by ip address

kimhanbeom:
- This fix ensures that `ExternalDdos`, `MultiHostPortScan`, and `RdpBruteForce` with multiple addresses are filtered correctly.
  - Change the return types of `src_addr` and `dst_addr` in the `Match` trait: `IpAddr` to `&[IpAddr]`.
  - Change the names of `src_addr` and `dst_addr` in the `Match` trait to plural: `src_addrs`, `dst_addrs`.

Close: #411

### DIFF
--- a/src/event/bootp.rs
+++ b/src/event/bootp.rs
@@ -136,16 +136,16 @@ impl BlockListBootp {
 }
 
 impl Match for BlockListBootp {
-    fn src_addr(&self) -> IpAddr {
-        self.src_addr
+    fn src_addrs(&self) -> &[IpAddr] {
+        std::slice::from_ref(&self.src_addr)
     }
 
     fn src_port(&self) -> u16 {
         self.src_port
     }
 
-    fn dst_addr(&self) -> IpAddr {
-        self.dst_addr
+    fn dst_addrs(&self) -> &[IpAddr] {
+        std::slice::from_ref(&self.dst_addr)
     }
 
     fn dst_port(&self) -> u16 {
--- a/src/event/common.rs
+++ b/src/event/common.rs
@@ -16,10 +16,10 @@ use super::{
 // TODO: Make new Match trait to support Windows Events
 
 pub(super) trait Match {
-    fn src_addr(&self) -> IpAddr;
+    fn src_addrs(&self) -> &[IpAddr];
     #[allow(dead_code)] // for future use
     fn src_port(&self) -> u16;
-    fn dst_addr(&self) -> IpAddr;
+    fn dst_addrs(&self) -> &[IpAddr];
     #[allow(dead_code)] // for future use
     fn dst_port(&self) -> u16;
     #[allow(dead_code)] // for future use
@@ -82,7 +82,13 @@ pub(super) trait Match {
     ) -> Result<(bool, Option<Vec<TriageScore>>)> {
         if let Some(customers) = &filter.customers {
             if customers.iter().all(|customer| {
-                !customer.contains(self.src_addr()) && !customer.contains(self.dst_addr())
+                self.src_addrs()
+                    .iter()
+                    .all(|&src_addr| !customer.contains(src_addr))
+                    && self
+                        .dst_addrs()
+                        .iter()
+                        .all(|&dst_addr| !customer.contains(dst_addr))
             }) {
                 return Ok((false, None));
             }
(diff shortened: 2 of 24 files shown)


src/event/bootp.rs
```diff
@@ -136,16 +136,16 @@ impl BlockListBootp {
 }
 
 impl Match for BlockListBootp {
-    fn src_addr(&self) -> IpAddr {
-        self.src_addr
+    fn src_addrs(&self) -> Vec<IpAddr> {
+        vec![self.src_addr]
```
sehkone, line 140: Since it always returns only a single IP address, returning a vector doesn’t seem like a good idea. It would be better to keep both `src_addr` and `src_addrs`, and use each according to the data type. To avoid confusion, using `src_addr_list` instead of `src_addr` might be a better option. This will require implementing `other_matches` specifically  for some types.

kimhanbeom: done

sehkone:
I suggest a new type for both single IP address and multiple ones as follows:

enum IpEntry {
    Single(IpAddr),
}

sehkone:
> This could reduce the complexity of maintaining logic to determine whether a single address or a list is appropriate in a given context.

I believe we can define the following methods in the `Match` trait.

```
   fn src_ip_entry(&self) -> IpEntry;
   fn dst_ip_entry(&self) -> IpEntry;
   fn other_matches(&self ..) -> ...
```

This way will not require an additional method, `multi_ip_address_matches`.

src/event/common.rs
```diff
@@ -142,13 +140,25 @@ pub(super) trait Match {
...
             } else {
-                return Ok((false, None));
+                return false;
             }
         }
 
+        true
+    }
+
+    /// Returns whether the event matches the filter (excluding `kinds`, ip address compare field)
```
sehkone, line 153: This comment is about the English. The phrase “excluding kinds, IP address compare field” seems to suggest that kinds is the IP address compare field itself. Using just a comma usually implies that the former and the latter are the same, so we need to explicitly add “and” to avoid confusion. For two items, we write “A and B”; for three items, we write “A, B, and C”.

sehkone:
@kimhanbeom 

I apologize for confusing you with my changing suggestions. I mistakenly thought the `Match` trait was one of the public interfaces that needed to distinguish between a single IP and multiple IPs. It turns out that it's internal and doesn’t require strict method separation. Therefore, your initial implementation should work well with a few alterations:

```
   fn src_addrs(&self) -> Vec<IpAddr> {
        vec![self.src_addr]
```

It would be better to modify the above to:

```
   fn src_addr_entry(&self) -> &[IpAddr] {
        std::slice::from_ref(&self.src_addr)
```

This should be more efficient by using a slice instead of a vector.

src/event/common.rs
kimhanbeom: done

kimhanbeom:
Restored the initial implementation, applied `src_addr_entry` and `dst_addr_entry`, and fixed Clippy errors introduced by updated Rust versions(1.86.0).

sehkone (approved):

sehkone:
@kimhanbeom, 

I think this needs rebasing.

msk:
> * **Naming:** The renaming of `src_addr` to `src_addr_entry` and `dst_addr` to `dst_addr_entry` is slightly unconventional. While the `_entry` suffix is sometimes used to indicate a method returning a reference to an internal field, it doesn't clearly convey the purpose of the method (i.e., providing the relevant IP addresses for matching, which could be multiple).  Alternatives like `src_addrs` and `dst_addrs` (as initially suggested in the PR description) might be more intuitive, even though they now return slices.

@kimhanbeom What do you think of this issue Octoaide pointed out?

sehkone:
> > * **Naming:** The renaming of `src_addr` to `src_addr_entry` and `dst_addr` to `dst_addr_entry` is slightly unconventional. While the `_entry` suffix is sometimes used to indicate a method returning a reference to an internal field, it doesn't clearly convey the purpose of the method (i.e., providing the relevant IP addresses for matching, which could be multiple).  Alternatives like `src_addrs` and `dst_addrs` (as initially suggested in the PR description) might be more intuitive, even though they now return slices.
> 
> @kimhanbeom What do you think of this issue Octoaide pointed out?

@kimhanbeom, I agree with what @msk suggested. What I emphasized in my previous comment is to use a slice, not the term 'entry'.

PR #430, opened 2025-04-28
Modify to create VERSION file if the file does not exist

### LINKED ISSUE #429: Modify to create VERSION file if the file does not exist
Related: https://github.com/aicers/codepot/issues/125

Currently the VERSION file of <data_dir> is created only if the <data_dir> path does not exist.
This procedure should be modified to create the VERSION file if it does not exist.

https://github.com/petabi/review-database/blob/5bd968a17154d2e529abe980390d152676fffa61/src/migration.rs#L244-L258

syncpark:
Close #429 

- Modify to create the VERSION file if the file does not exist, not if the "<data_dir>" path does not exist.


syncpark:
I will close this issue as it's decided to address it in a different way.

PR #447, opened 2025-06-06
Modify `Node` structure

### LINKED ISSUE #399: Modify the node structure to support Mitre infomation module

sehkone:
This PR originates from #401 and consists solely of renaming `Remote`-related items to `UnlinkedServer`-related ones—no other changes were made.

@msk, I made this adjustment based on your feedback regarding the term `Remote`. Although `UnlinkedServer` is a bit longer, it seems to better reflect the intended meaning.

If you’re okay with this naming, could you consider this PR as the replacement for #401?

Close #399

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,15 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   naming convention used in other parts of the codebase.
 - Modified the `ValueKind` enum to support different types of input for packet
   attribute triage.
+- Modified the `Node` structure to allow you to manage the configuration of all
+  remote servers that communicate with the REview.
+  - Introduced `Remote`, a new structure that stores only configuration drafts
+    for remote servers.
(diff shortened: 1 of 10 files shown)


sehkone:
When creating this PR, I initially thought it would be enough to tinker with what @kimhanbeom uploaded in #401. However, I’ve identified some critical issues that require more substantial changes:

- `AgentKind` and `UnlinkedServerKind` currently share the same variants, but they shouldn’t. Each should define its own variants separately, as their internal members differ.
- We had previously decided to rename `Datalake` to `DataStore`, but it looks like that change was missed in #401.

@msk, I’m closing this PR and asking @kimhanbeom to incorporate the updates made here and also address the above issues.

Issue #548, opened 2025-08-29
Incorrect Struct Usage in Detection Event Migration


kimhanbeom:
## Background
Some detection event migration functions in review-database currently use incorrect struct definitions during deserialization and serialization.

## Issue Detail
- The semi-supervised engine serializes detection events using structs that do not include the `time` and `triage_scores` fields, and sends them to REview. These events are then stored directly in the database.
   ```rust 
    // Example: `RdpBruteForce` detection event
    struct RdpBruteForceFields {
        src_addr: IpAddr,
        dst_addrs: Vec<IpAddr>,
        start_time: DateTime<Utc>,
        end_time: DateTime<Utc>,
        proto: u8,
        confidence: f32,
        category: EventCategory,
    }
   ```
- However, the current migration code assumes that all structs used for deserialization and serialization contain `time` and `triage_scores`.
- As a result, deserialization fails, preventing proper migration.

## Task
- To resolve this issue, migration logic must deserialize using the struct format that matches the data stored in the database, and serialize using the latest struct format. For example, the correct migration for the `RdpBruteForce` detection event should look like this:
   ```rust
    // Old structure (we should define in `src/migration/migration_structures.rs`)
    #[derive(Debug, Deserialize, Serialize)]
    pub struct RdpBruteForceV0_39 {
        pub src_addr: IpAddr,
        pub dst_addrs: Vec<IpAddr>,
        pub start_time: DateTime<Utc>,
        pub end_time: DateTime<Utc>,
        pub proto: u8,
        pub category: EventCategory,
    }
    // latest structure (defined in `src/event/rdp.rs`)
    #[derive(Serialize, Deserialize)]
    pub struct RdpBruteForceFields {
        pub src_addr: IpAddr,
        pub dst_addrs: Vec<IpAddr>,
        pub start_time: DateTime<Utc>,
        pub end_time: DateTime<Utc>,
        pub proto: u8,
        pub confidence: f32,
        pub category: EventCategory,
    }
    // From Trait Implement (we should Implement in `src/migration/migration_structures.rs`)
    impl From<RdpBruteForceV0_39> for RdpBruteForceFields {
        fn from(old: RdpBruteForceV0_39) -> Self {
            Self {
                src_addr: old.src_addr,
                dst_addrs: old.dst_addrs,
                start_time: old.start_time,
                end_time: old.end_time,
                proto: old.proto,
                confidence: 0.3,
                category: old.category,
            }
        }
    }
[...]

PR #549, opened 2025-09-01
Fix event migration

msk:
This should fix logic errors in event migration reported in #548.


kimhanbeom:
@msk 
Everything looks great overall. I just have one suggestion — it would be preferable to keep the migration-related structs and `From` implementations in "migration_structure.rs", as we've done previously.

msk:
@kimhanbeom Thanks for the review!

I understand your preference for keeping migration structures centralized, and I agree that was the cleaner approach historically. However, the new pattern I've implemented addresses a practical challenge with AI coding agents.

The Problem with the previous approach is that, in the old pattern, when an AI agent needed to add a field to a struct (like adding confidence to event structures), it had to:

1. Copy the current struct definition to migration_structures.rs with a versioned name
2. Modify the original struct in place
3. Implement the From trait between old and new versions
4. Update migration functions

This multi-step, cross-file process frequently caused AI agents to miss steps or implement incomplete migrations.

The new pattern uses versioned structs directly in their domain modules with type aliases:

- `FtpBruteForceFieldsV0_39` (old version)
- `FtpBruteForceFieldsV0_41` (new version)
- `pub type FtpBruteForceFields = FtpBruteForceFieldsV0_41;` (current alias)

I believe this would work better for AI because:

1. Single-file changes: When a struct needs modification, the AI only adds a new versioned struct and updates one type alias in the same file
2. No cross-module dependencies: Everything needed is co-located
3. Clearer versioning: The active version is explicitly shown via the type alias
4. Simpler mental model: Add new version + update alias, rather than copy-modify-implement pattern

The versioned structs in individual modules (like `FtpBruteForceFieldsV0_39`) serve the same purpose as the old migration_structures.rs; they preserve the old format for migration while making the process more AI-friendly.

This approach maintains the same migration capabilities while reducing the cognitive load and error potential for AI coding agents. What do you think of this trade-off?

kimhanbeom:
@msk 
I agree — the approach you've suggested makes sense for improving code productivity through AI.

kimhanbeom:
@msk 
One additional change is required: please exclude migration handling for the `FtpPlainText` and `LdapPlainText` events.

Migration for these two events was originally planned for v41 due to the addition of the `confidence` field. However, the semi-supervised engine has already been sending these events with the `confidence` field included since v39. As a result, their structure is already consistent with the current version, and migration is no longer necessary. These events should therefore be excluded from the migration process.

PR #616, opened 2025-11-05
Simplify key settings

minshao:
I've modified the key according to my comments. Could you show me how it would cause a Seized compilation error?


syncpark (commented):
It seems like the problem occurred before adding the Sealed trait in "types.rs".

msk:
The base branch has been merged to main.

Issue #658, opened 2025-12-11
`migrate_backend` is not idempotent and fails on re-run, on `stable-0.42`


sophie-cluml:
### Summary

On the stable-0.42 branch tip, in environments where PostgreSQL already contains model and/or cluster data to migrate, running `review-migrate` succeeds on the first run but fails on the second run.

The root cause appears to be that two steps inside `migrate_backend` exhibit non-idempotent behavior:

1. `update_cluster_id_in_column_stats` → fails with `Error: Unable to find Cluster id`
2. `insert_models` → fails with `Error: Model map is not empty. Aborting migration.`

Although the second failure only appears after the first one is bypassed or otherwise allowed to succeed, the underlying issue is that `migrate_backend` itself is not idempotent. As a result, the backend migration pipeline cannot be safely retried by binaries that use the `migrate_backend` public API.

### Steps to Reproduce

1. First run -  migration succeeds

```
$ git checkout origin/stable-0.42
$ cargo run --bin review-migrate --features=migrate review-migrate.toml  # NOTE: review-migrate.toml is just a configuration file
Starting migration process...
Migrating data directory...
Connecting to PostgreSQL database...
Migrating data from PostgreSQL to RocksDB...
Migration completed successfully.
```

2. Second run — `update_cluster_id_in_column_stats` fails

Running the same command again without changing anything:

```
$ cargo run --bin review-migrate --features=migrate review-migrate.toml
Starting migration process...
Migrating data directory...
Connecting to PostgreSQL database...
Migrating data from PostgreSQL to RocksDB...
Error: Unable to find Cluster id
```

3. After bypassing the above error — `insert_models` fails

Once the above-failing cluster migration step is bypassed or made to succeed (even trivially), the next failure occurs:

```
Starting migration process...
Migrating data directory...
Connecting to PostgreSQL database...
Migrating data from PostgreSQL to RocksDB...
Error: Model map is not empty. Aborting migration.
```

### Observed Behavior in the Two Functions

- On `update_cluster_id_in_column_stats`
  - The function assumes that all column-stats keys are still in the old V41 format.
  - After the first migration run, these keys have already been rewritten into the new format, and the corresponding rows in the PostgreSQL `cluster` table have been removed via `remove_clusters`.
  - A second run therefore attempts to interpret *new-format* keys as if they were *old-format* keys.
  - The derived `cluster_id` no longer exists in PostgreSQL, resulting in `Unable to find Cluster id`.
- On `insert_models`
[...]

PR #659, opened 2025-12-11
Make backend migration idempotent

sophie-cluml:
### Summary

Skip migration steps if source data in PostgreSQL is already removed.

Close: #658

-- 

### Detailed Description

This PR addresses issue #658, where backend migration fails on subsequent runs even though the first migration completes successfully.

In the current design, the backend migration (`migrate_backend`) removes source data from PostgreSQL after each component successfully migrates its data into RocksDB. As a result, on subsequent runs, an empty PostgreSQL table may legitimately indicate either:

(a) the migration step was already completed in a previous run, or
(b) there was nothing to migrate in the first place.

Prior to this PR, two components did not handle this scenario correctly:
- `update_cluster_id_in_column_stats` attempted to run even when the PostgreSQL cluster table was already empty, causing it to misinterpret already-migrated RocksDB keys as old-format keys and fail.
- `insert_models` had a different failure mode: even when PostgreSQL contained no models to migrate, the function still treated the presence of existing model entries in RocksDB as an error condition, unconditionally aborting the migration. This did not allow any migration re-runs.

This PR ensures idempotency by explicitly skipping these migration steps when the corresponding PostgreSQL source data is empty.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -621,8 +621,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   generating `rustls::ClientConfig`.
 - Changed the display message format of `EventMessage` and `Event` to RFC 5424.
   Modified messages will be sent to syslog.
-- Used `FromPrimitive` and `ToPrimitive` for converting `EventCategory` instead of
-  manually implementing `TryFrom`.
+- Used `FromPrimitive` and `ToPrimitive` for converting `EventCategory` instead
+  of manually implementing `TryFrom`.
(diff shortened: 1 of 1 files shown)


sophie-cluml:
I manually tested the changes by running both the initial migration and subsequent re-runs under two conditions: (1) an empty PostgreSQL and RocksDB database (2) a database populated with cluster, model, and column statistics data. Both scenarios succeeded, for example like below:

image

sehkone (approved):

PR #728, opened 2026-05-08
Add inline exclusion/policy matching helpers (#724)

### LINKED ISSUE #724: Add public helpers for inline exclusion/policy matching (1B-4b, database-side)
Add two small public helpers to `review-database` so `review-web`'s new `eventListWithTriage` resolver (aicers/review-web#842) can perform inline exclusion matching and inline policy scoring without widening the visibility of the `Match` trait itself.

## Helpers to add

```rust
impl Event {
    /// Returns true if any of `exclusions` matches this event.
    /// Uses the existing per-event `score_by_triage_exclusion` overrides.
    pub fn matches_exclusion(&self, exclusions: &[TriageExclusion]) -> bool { ... }

    /// Computes triage scores for this event against each of `policies`.
    /// Returns scores that exceed each policy's `minimum_score` threshold.
    pub fn score_against_policies(&self, policies: &[TriagePolicyInput]) -> Vec<TriageScore> { ... }
}
```

(Exact names and signatures are open to PR review — names listed here are illustrative. `matches_any_exclusion` / `score_with_policies` are reasonable alternatives.)

## Layering boundary

These helpers operate on `review-database`'s own runtime types — `Event`, `TriageExclusion`, `TriagePolicyInput`, `TriageScore`. **The helpers do NOT know about any GraphQL input type**: `EventTriagePolicyInput`, `EventTriageExclusionInput`, `EventTriageInput` live in `review-web` only. The conversion `EventTriagePolicyInput` → `TriagePolicyInput` happens entirely in `review-web` (see aicers/review-web#842 step 3) and the helpers receive only the converted slice.

This preserves the existing `review-database` ↔ `review-web` boundary: GraphQL types and the matching engine remain in their respective crates without bleeding into each other.

## Implementation notes

- `Event` enum is already `pub`; the existing `Event::matches` ([src/event.rs:513-565](https://github.com/aicers/review-database/blob/main/src/event.rs)) shows the dispatch pattern to follow — one `match self { ... }` arm per event variant, including the `Blocklist(RecordType)` sub-dispatch.
- `Match` trait visibility is **not** widened. Only the two function-shaped helpers become `pub`. Internal trait stays `pub(super) trait Match`.
- Both helpers reuse the existing `score_by_attr`, `score_by_confidence`, `score_by_triage_exclusion` dispatch (default impls plus DNS/HTTP/Tor per-event overrides) — no new matching algorithm.
- **Avoid duplicating threshold/`TriageScore`-construction logic, but keep legacy and inline score formulas separate.** The existing `Match::other_matches` computes `score_by_triage_exclusion + score_by_attr + score_by_confidence` and threshold-filters by `response.minimum_score` ([src/event/common.rs:248-271](https://github.com/aicers/review-database/blob/main/src/event/common.rs)). The new inline helper uses a different formula: `score_by_attr + score_by_confidence` only, because the resolver applied exclusions in step 2 (see aicers/review-web#842). Therefore:
  - Extract only the **shared** part — applying `response.minimum_score` and constructing `TriageScore { policy_id, score }` — into a small private helper (e.g. `build_triage_score(policy_id, score, response) -> Option<TriageScore>`).
  - Keep the per-policy score *calculation* separate: the legacy path calls all three `score_by_*`, the inline path calls only `score_by_attr` + `score_by_confidence`. Do **not** introduce a single `score_one_policy` that both paths share, since that would force the legacy formula onto the inline path or vice versa.
- `score_against_policies` interprets each `TriagePolicyInput.triage_exclusion` as "already empty" (the resolver applied exclusions at step 2), so it only invokes `score_by_attr` and `score_by_confidence`, then filters by `response.minimum_score`. **Enforce this contract** with a `debug_assert!(p.triage_exclusion.is_empty())` so accidentally-populated exclusions in the input are caught in development rather than being silently ignored.
- When `response` is empty for a given policy, the score is dropped (matches existing `other_matches` behavior). Document this on the public helper so callers don't expect "no threshold = always score".
- Exclusion matching: the **default** `score_by_triage_exclusion` impl in [src/event/common.rs:276-286](https://github.com/aicers/review-database/blob/main/src/event/common.rs) only handles the `IpAddress` variant. Per-event overrides already cover richer shapes — DNS variants match `Domain`, HTTP variants match `Uri`/`Hostname`, Tor matches `Domain` — and `matches_exclusion` must dispatch through these overrides exactly like `Event::matches` does, so the existing override semantics are preserved. If aice-web-next#457 (1B-2) introduces *new* exclusion shapes beyond what the current overrides cover, extending the matchers is a follow-up issue and **not** part of this scope.
- `matches_exclusion` is implemented on top of `score_by_triage_exclusion`, which uses `f64::MIN` as a sentinel for "matched" ([src/event/common.rs:285](https://github.com/aicers/review-database/blob/main/src/event/common.rs)). Hide this internal contract behind a small private helper (e.g. `fn matched_any_exclusion(&self, exclusions: &[TriageExclusion]) -> bool { self.score_by_triage_exclusion(exclusions) == f64::MIN }`) inside the `Match` trait so the sentinel does not leak into the public helper's body. Do not expose `f64::MIN` comparisons in the public-facing code.
- The existing `TriagePolicyInput.id: u32` and `TriageScore.policy_id: u32` types are unchanged. Callers (aicers/review-web#842) supply integer IDs through the GraphQL surface (`Int!`, not `ID!`) so the round-trip identity holds without any string mapping.
- Code is roughly 100–150 lines of dispatch boilerplate plus tests. No new algorithm, no new type.

## Required tests

At minimum the PR must cover:

- One representative top-level `Event` variant (e.g. `DnsCovertChannel`) — exclusion match and non-match.
- One `Blocklist(RecordType)` sub-variant (e.g. `RecordType::Http`) — to exercise the nested dispatch.
- `score_against_policies` with an empty `policies` slice → returns empty `Vec`.
- `score_against_policies` where one policy passes `minimum_score` and one fails → only the passing score is returned.
- A policy with empty `response` → that policy contributes no score (documents existing semantics).
- **Legacy regression**: `Event::matches(..., filter.triage_policies = Some(non-empty))` still drops the event when no policy's score reaches `minimum_score`. This guards the shared-helper refactor against accidentally changing legacy semantics — the legacy path must keep using `score_by_triage_exclusion + score_by_attr + score_by_confidence`.

## Add-only

The `Match` trait's **visibility and existing behavior are unchanged** — `score_by_*` defaults, all per-event overrides, `EventDb::put`, and all matching code keep their current semantics. The trait body itself gains a small private default method (`build_triage_score`, `matched_any_exclusion`) as described in Implementation notes, but no existing method or its observable behavior is modified. The two new public helpers on `Event` are pure additions and are backward compatible.

## Required by

- aicers/review-web#842.

## Discussion

See https://github.com/aicers/aice-web-next/discussions/447 §2.2.

sehkone:
## Summary

Adds two public helpers on `Event` so `review-web`'s upcoming `eventListWithTriage` resolver can perform inline triage exclusion matching and inline policy scoring without widening the visibility of the internal `Match` trait:

- `Event::matches_exclusion(&[TriageExclusion]) -> bool` — dispatches through each event variant's `score_by_triage_exclusion` override (including the `Blocklist(RecordType)` sub-dispatch), preserving DNS/HTTP/Tor/etc. per-event semantics. The `f64::MIN` sentinel used internally by `score_by_triage_exclusion` is hidden behind a private `Match::matched_any_exclusion` helper so it does not leak into public code.
- `Event::score_against_policies(&[TriagePolicyInput]) -> Vec<TriageScore>` — sums only `score_by_attr + score_by_confidence` (the resolver is expected to apply exclusions in a prior step), then threshold-filters by `response.minimum_score`. A `debug_assert!` enforces that each input policy's `triage_exclusion` is empty. Policies with an empty `response` contribute no score, matching existing `other_matches` behavior.

Shared `TriageScore` construction (`Match::build_triage_score`) is factored out so the legacy and inline paths stay in sync without coupling their score formulas. The legacy `Match::other_matches` formula is unchanged. The `Match` trait's visibility is not widened; only the two new helpers on `Event` become public.

Closes #724

## Test plan

- [x] `matches_exclusion` matches and non-matches for a top-level variant (`DnsCovertChannel` via `Hostname`)
- [x] `matches_exclusion` exercises the `Blocklist(RecordType)` nested dispatch (`Blocklist::Http` via `Uri`)
- [x] `score_against_policies` with an empty `policies` slice returns an empty `Vec`
- [x] `score_against_policies` returns only the policies whose score reaches `minimum_score`
- [x] `score_against_policies` drops policies whose `response` is empty
- [x] Legacy regression: `Event::matches` still drops the event when no policy's score reaches `minimum_score` (guards the shared-helper refactor)
- [x] `cargo fmt -- --check --config group_imports=StdExternalCrate`
- [x] `cargo clippy --bins --tests --all-features -- -D warnings`
- [x] `cargo test --all-features` (new event::common tests pass)
- [x] `markdownlint-cli2` on `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   `clear_retention_config` methods on `Store` for managing the retention
   period. Added `purge_old_column_stats` method to `Store` that deletes column
   statistics older than the configured retention period.
+- Added `Event::matches_exclusion` and `Event::score_against_policies` public
+  helpers so callers can perform inline triage exclusion matching and inline
+  policy scoring without going through the full `Event::matches` filter
+  pipeline. `score_against_policies` sums only `score_by_attr +
+  score_by_confidence` and assumes exclusions have already been applied by the
+  caller; policies whose `response` is empty contribute no score.
 
 ### Changed
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -84,7 +84,7 @@ pub use self::{
     unusual_destination_pattern::{UnusualDestinationPattern, UnusualDestinationPatternFields},
 };
 use super::{
-    Customer, EventCategory, Network, TriagePolicyInput,
+    Customer, EventCategory, Network, TriageExclusion, TriagePolicyInput,
     types::{Endpoint, HostNetworkGroup},
 };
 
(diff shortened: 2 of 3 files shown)


sehkone:
**[Reviewer Round 1]**

No blocking findings.

I reviewed this against #724 and the implementation looks aligned with the requested database-side helper boundary: `src/event.rs:567` adds the public `Event::matches_exclusion` dispatch without widening `Match`, and `src/event.rs:625` adds `Event::score_against_policies` on the already-public `Event` type. The scoring split is also preserved: the legacy `Event::matches` path still computes `score_by_triage_exclusion + score_by_attr + score_by_confidence` before thresholding (`src/event/common.rs:250`), while inline scoring uses only `score_by_attr + score_by_confidence` and debug-asserts that exclusions were already applied (`src/event/common.rs:320`). The shared threshold/`TriageScore` construction at `src/event/common.rs:297` keeps the empty-response behavior consistent without coupling the legacy and inline formulas.

The tests cover the required surface in a meaningful way: top-level exclusion match/non-match, `Blocklist(Http)` nested dispatch, empty policy slices, pass/fail threshold filtering, empty `response` dropping, and a legacy no-policy-passes regression (`src/event/common.rs:2890`, `src/event/common.rs:2906`, `src/event/common.rs:2918`, `src/event/common.rs:2924`, `src/event/common.rs:2947`, `src/event/common.rs:2960`). PR hygiene also looks good: the body uses `Closes #724`, includes a checked test plan, and `CHANGELOG.md:25` documents the new public helpers and inline scoring contract.

sehkone:

PR #729, opened 2026-05-09
Make node Table::update fully atomic and return final Node (#721)

### LINKED ISSUE #721: Make node Table::update fully atomic and return the post-update Node
### Background

`aicers/review-web` is splitting its `applyNode` mutation (see [discussions/548](https://github.com/aicers/review-web/discussions/548)). To support that split, `Table::update` in this repository needs two properties it currently lacks:

1. **True atomicity across the Node / Agent / ExternalService tables.** Today the node's inner record is updated inside an optimistic transaction, but the agent and external-service table updates happen outside it. A partial-update state (node inner record updated, agent/external_service updates partially applied) is therefore observable in principle.
2. **An outcome returned from the update call so callers do not need a separate read-back.** Today `Table::update` returns `()`, so callers must read the node back to learn its post-update state. That post-commit read reopens a write→read gap that is one of the race concerns raised in #548, and should be closed at the DB layer rather than papered over above it.

Hostname uniqueness at the DB layer, raised by the earlier issue [#496](https://github.com/aicers/review-database/issues/496), has already landed (0.40.0, 2025-08-13). This issue builds on top of that and does not revisit it.

### Current state

`src/tables/node.rs`:

```rust
pub fn update(&mut self, id: u32, old: &Update, new: &Update) -> Result<()>
```

- The node inner record update runs inside an optimistic transaction with a hostname-uniqueness check (lines 456–529 at the time of writing).
- **Agent table operations** (`put` / `delete` / `update`) run **outside** that transaction (lines 531–564).
- **ExternalService table operations** likewise run **outside** the transaction (lines 566–604).
- The function returns `()`, forcing callers to read back the node if they need the post-update state.

Note that the generic `Table<R>` already exposes transaction-aware `put_with_transaction`, `insert_with_transaction`, `update_with_transaction`, and `delete_with_transaction` (see `src/tables.rs`). Change 1 below does not need new primitives; it just requires using the existing ones.

### Proposed changes

The two changes below modify `Table::update` in `src/tables/node.rs`. They are delivered together by changing the existing function's internal behavior and return type — **not** by introducing a parallel method. This is a **BREAKING** change to the public signature and to the implicit partial-atomicity that the current implementation exhibits.

#### Change 1 — Atomicity across Node / Agent / ExternalService

`Table::update` must perform all Node, Agent, and ExternalService write operations inside a single transaction. Either the entire update succeeds or none of its effects become visible.

No public API additions are required for this: the necessary transaction-aware operations (`*_with_transaction`) already exist on `Table<R>`. The change is internal to `Table::update`'s implementation.

#### Change 2 — The node-update path returns the final `Node`

The node-update path must return enough information that the caller does not need a separate `get_by_id` read after the call. The only piece of state the caller cannot reconstruct on its own is the final `Node` with its preserved `creation_time` (which is not part of the `Update` input). Therefore `Table::update`'s return type changes from `Result<()>` to `Result<Node>`. This is a **BREAKING** change.

How the returned `Node` is constructed:

- The transaction already reads the current `Inner` record while performing the update, so `creation_time` is available without a second read.
- The post-commit `Node` is assembled inside the transaction from: the existing `Inner` record (for `creation_time`); the applied `new` update (for `name`, `name_draft`, `profile`, `profile_draft`); and `new.agents` / `new.external_services` with each element's `node` field normalized to the node's `id` — the same normalization the write path already performs immediately before each agent/external_service write (see the current `update` implementation). No post-commit `get_by_id` is needed.

Record the signature change in `CHANGELOG.md` per this repo's existing BREAKING-change conventions (prose-only, listing the affected API and its before/after).

### Acceptance criteria

- [ ] `Table::update` performs its Node, Agent, and ExternalService write operations inside a single transaction.
- [ ] If any pre-commit step of `Table::update` fails, no database state change is observable afterward. The specific failure-induction mechanism is at the implementer's discretion — for example, supply an `Update` whose agent list forces an error at the agent-write step after the node-inner-write step has already staged its change within the transaction, and verify that the committed state shows neither change. At least one such end-to-end scenario must be covered.
- [ ] `Table::update` returns the post-update `Node` — agents and external_services included, `creation_time` preserved from the existing record — without performing a post-commit `get_by_id` read.
- [ ] Unit tests cover, at minimum, for the returned `Node`:
  - [ ] agent added — the new agent appears in `returned_node.agents`
  - [ ] agent removed — it is absent from `returned_node.agents`
  - [ ] profile change — `returned_node.profile` reflects the applied change
  - [ ] `creation_time` is preserved from the pre-update record
  - [ ] no-op update — the returned `Node` reflects the unchanged state
- [ ] `CHANGELOG.md` updated under **Changed** / **BREAKING** with a prose description of the signature change, matching existing entries in this repo.

### CHANGELOG notes

- `Table::update` is now fully atomic across the Node, Agent, and ExternalService tables. Callers that relied on observing partial-update intermediate states (none should) are no longer supported.
- `Table::update`'s return type changed from `Result<()>` to `Result<Node>`. Describe the before/after in prose, following existing BREAKING entries in this CHANGELOG.

### References

- `aicers/review-web` discussion: [#548](https://github.com/aicers/review-web/discussions/548)
- Earlier DB-layer work (closed): [#496 Add Uniqueness Constraint for Hostnames](https://github.com/aicers/review-database/issues/496)
- Current implementation:
  - `src/tables/node.rs` — `Table::update`, `Table::put`, `Table::get_by_id`
  - `src/tables.rs` — generic `Table<R>` with its `*_with_transaction` primitives already in place
  - `src/tables/agent.rs` — `Agent` record
  - `src/tables/external_service.rs` — `ExternalService` record

sehkone:
## Summary

- Fold all Node, Agent, and ExternalService write operations in `tables::node::Table::update` into a single optimistic transaction so either the entire update commits or no state change becomes visible. Previously, only the node's inner record was updated within a transaction, while the agent and external-service writes ran outside it, making partial-update intermediate states observable in principle.
- Change the return type of `Table::update` from `Result<()>` to `Result<Node>`. The returned `Node` is assembled inside the successfully-committed transaction attempt: `creation_time` comes from the existing record (read under the same snapshot via the new `get_by_id_in_transaction` primitive); `name`, `name_draft`, `profile`, and `profile_draft` come from the applied update; `agents` and `external_services` come from `new.agents` / `new.external_services` with each element's `node` field normalized to the node's `id`. Callers no longer need a follow-up `get_by_id` read.
- Each retry iteration of the optimistic-commit loop reads its own snapshot and constructs its own candidate `Node`; only the iteration whose `txn.commit()` returns `Ok` returns its value, so data observed in failed/retried attempts is never reused.
- Add `IndexedTable::get_by_id_in_transaction` (and the underlying `Indexed::get_by_id_in_transaction`) so the node update can read the existing inner record under the same transaction with an exclusive lock.
- Document both behavioral changes as BREAKING in `CHANGELOG.md`.

Both changes are BREAKING, both for the implicit partial-atomicity that the previous implementation exhibited and for the public signature of `Table::update`.

Closes #721

## Test plan

- [x] `cargo fmt -- --check --config group_imports=StdExternalCrate` passes.
- [x] `cargo clippy --bins --tests --all-features -- -D warnings` passes.
- [x] `markdownlint-cli2 CHANGELOG.md` passes.
- [x] `cargo test --all-features tables::node` — all 18 tests pass, including:
  - [x] `update_returns_node_with_added_agent` — added agent appears in the returned `Node.agents`.
  - [x] `update_returns_node_with_removed_agent` — removed agent is absent from the returned `Node.agents`.
  - [x] `update_returns_node_with_changed_profile` — `Node.profile` reflects the applied change.
  - [x] `update_preserves_creation_time` — `creation_time` is preserved from the pre-update record.
  - [x] `update_noop_returns_unchanged_node` — a no-op update returns the unchanged `Node`.
  - [x] `update_atomic_rollback_on_agent_failure` — when an inner step fails after the node-inner write has staged its change within the transaction, no observable state change remains afterward (verifies cross-table atomicity).
- [x] CI `check`, `test` (ubuntu-latest, macOS-latest), and `coverage` jobs are green on this branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,23 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   category is absent; `Some(category)` preserves existing
   behavior. A migration converts persisted records so previous
   category values become `Some(...)`.
+- **BREAKING**: `tables::node::Table::update` is now fully atomic
+  across the Node, Agent, and ExternalService tables. All Node,
+  Agent, and ExternalService write operations execute inside a
+  single optimistic transaction, so either the entire update
+  commits or no state change is observable. Callers that relied
+  on observing intermediate partial-update states are no longer
+  supported.
+- **BREAKING**: `tables::node::Table::update`'s return type changed
+  from `Result<()>` to `Result<Node>`. The returned `Node` is
+  assembled inside the successfully-committed transaction
+  attempt: `creation_time` comes from the existing record;
+  `name`, `name_draft`, `profile`, and `profile_draft` come from
+  the applied update; and `agents` / `external_services` come
+  from `new.agents` / `new.external_services` with each element's
+  `node` field normalized to the node's `id`. Callers no longer
+  need to issue a follow-up `get_by_id` read to observe the
+  post-update state.
 
 ## [0.44.1] - 2026-04-16
 
--- a/src/collections.rs
+++ b/src/collections.rs
@@ -313,6 +313,28 @@ pub trait Indexed {
             .transpose()
     }
 
+    /// Gets an entry corresponding to the given index within a transaction,
+    /// acquiring an exclusive lock on the entry.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the index is invalid or cannot be read.
+    fn get_by_id_in_transaction<T: Indexable + FromKeyValue>(
+        &self,
+        id: u32,
+        txn: &rocksdb::Transaction<rocksdb::OptimisticTransactionDB>,
+    ) -> Result<Option<T>> {
+        let index = self.index_in_transaction(txn)?;
+        let Some(key) = index.get(id).context("invalid ID")? else {
+            return Ok(None);
+        };
+        let key = T::make_indexed_key(Cow::Borrowed(key), id);
+        txn.get_for_update_cf(self.cf(), &key, super::EXCLUSIVE)
+            .context("cannot read entry")?
+            .map(|value| T::from_key_value(&key, &value))
+            .transpose()
+    }
+
     /// Inserts a new key-value pair.
     ///
     /// # Errors
--- a/src/tables.rs
+++ b/src/tables.rs
@@ -820,6 +820,23 @@ impl<'d, R> IndexedTable<'d, R> {
         self.indexed_map.get_by_id(id)
     }
 
+    /// Gets a record with the given ID within a transaction, acquiring an
+    /// exclusive lock on the entry.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the database operation fails.
+    pub fn get_by_id_in_transaction(
+        &self,
+        id: u32,
+        txn: &rocksdb::Transaction<rocksdb::OptimisticTransactionDB>,
+    ) -> Result<Option<R>>
+    where
+        R: Indexable + FromKeyValue,
+    {
+        self.indexed_map.get_by_id_in_transaction(id, txn)
+    }
+
     /// Deactivates a key-value pair with the given ID.
     ///
     /// # Errors
--- a/src/tables/node.rs
+++ b/src/tables/node.rs
@@ -447,16 +447,47 @@ impl<'d> Table<'d> {
     ///  in both `old.agents` and `new.agents`, as well as each external service in both
     /// `old.external_services` and `new.external_services`, will be disregarded.
     ///
+    /// All `Node`, `Agent`, and `ExternalService` writes execute inside a single optimistic
+    /// transaction; either the entire update commits or no state change is observable.
+    /// Returns the post-update `Node` constructed from the successfully-committed
+    /// transaction attempt, with `creation_time` preserved from the existing record.
+    ///
     /// # Errors
     ///
     /// Returns an error if the `id` is invalid, the database operation fails, or if the hostname is already in use.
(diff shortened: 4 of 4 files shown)


sehkone:
**[Reviewer Round 1]**

No blocking findings.

I reviewed this against #721 and the implementation matches the core requirements: `Table::update` now stages the node inner update, agent diffs, and external-service diffs in the same optimistic transaction before a single commit (`src/tables/node.rs:487` through `src/tables/node.rs:643`). On any pre-commit error, the transaction is dropped before commit, so the wider write set should roll back as intended.

The returned `Node` is also built from the successful transaction attempt: the existing inner record is read under the transaction for `creation_time` (`src/tables/node.rs:513`), and the return value is only assembled on the `txn.commit()` success branch with agent and external-service `node` fields normalized to the update id (`src/tables/node.rs:643` through `src/tables/node.rs:677`). That avoids the post-commit `get_by_id` read the issue wanted to eliminate.

The test additions cover the required returned-node cases and include a meaningful rollback scenario where the node-inner write has already been staged before an agent compare/update failure (`src/tables/node.rs:1901` through `src/tables/node.rs:1961`). The changelog and PR hygiene also look aligned with the issue: the breaking return type and atomicity behavior are documented, the PR uses `Closes #721`, and the test plan is explicit.

sehkone:

PR #845, opened 2026-08-09
Set the classifier temp file's mode and flush it before rename (#844)

### LINKED ISSUE #844: Set the classifier temp file's mode and flush it before rename
# Set the classifier temp file's mode and flush it before rename

## Context

`AGENTS.md` requires two things of a temp-file-and-rename write. The temporary file is created with the permissions the finished file needs, because `rename` installs the temporary's inode at the destination and the destination therefore ends up with the temporary's mode, not the mode of the file it replaced. And atomic replacement is separated from durability: a file the program reads back to resume from needs `sync_all` on the temporary before the rename.

`ClassifierFileManager::store_classifier` in `src/classifier_fs.rs` does neither. It writes `{base_dir}/classifiers/model_{id}/classifier_{name}.bin` by writing a sibling temporary named with a millisecond timestamp extension and renaming it into place:

```rust
let timestamp = chrono::Utc::now().timestamp_millis();
let temp_path = file_path.with_extension(timestamp.to_string());

// Write to temporary file first
fs::write(&temp_path, data)                                    // no mode — the umask alone decides
    .map_err(|err| ClassifierFsError::FileWrite(err, temp_path.clone()))?;

// Rename to final location
if let Err(rename_err) = fs::rename(&temp_path, &file_path) {  // nothing flushed
```

**Permissions.** `fs::write` asks for `0o666` and lets the umask subtract from it, so a stored classifier ends up with whatever the umask happened to leave — under the common `umask 022` that is `0o644`, readable by every account on the host. Nobody chose that, and it is not necessarily the same twice if the umask differs between the process that first wrote the file and the one that later replaced it. Whatever mode the destination had before, including one an operator set by hand, is discarded at the next update.

Asking for `0o600` at creation instead is what this issue changes. Note what that does and does not guarantee, because the two are easy to conflate. A umask can only clear permission bits, never add them, so requesting `0o600` guarantees under **every** umask that no group or other bit is ever granted — that is the property worth having and the one worth testing. It does not pin the mode to exactly `0o600`: the umask still applies to the owner bits too, so a deployment running under `umask 0277` gets `0o400`. That is an operator narrowing their own files further, which is fine and needs no defending against.

**Durability.** The classifier is written once and read back by later runs, and the write is half of a pair. `upsert_model` (`src/lib.rs`) commits the model row to RocksDB first and only then calls `store_classifier`; `load_model_by_name` needs both halves and rejects the model when the file is missing. Temp-file-and-rename settles only which of two versions a reader sees; it says nothing about either surviving a power loss. Losing the unflushed write leaves the database listing a model whose file is unusable, in one of two ways:

- The rename never reaches disk, the file is absent, and every later `load_model_by_name` fails with `Classifier file not found for model {name}`.
- The rename reaches disk ahead of the data blocks, and the file is present but zero-length or partial. `classifier_exists` is satisfied by its presence and `load_classifier` returns the truncated bytes with no error, so the corrupt classifier is handed to the caller silently. This is the worse of the two and the reason the flush belongs here.

## Scope

In `store_classifier` in `src/classifier_fs.rs`, replace the `fs::write` call with an explicit create-write-flush sequence, keeping the surrounding structure as it is:

- Create the temporary through `std::fs::OpenOptions`, requesting `.mode(0o600)` via `std::os::unix::fs::OpenOptionsExt`, with `.write(true).create_new(true)`.
- Write `data` to it.
- Call `sync_all` on the file before the rename.

`create_new` rather than `create` is load-bearing here, not a style choice: a requested mode is applied only when `open` actually creates the file, so opening a temporary that already exists would silently keep that file's old mode and defeat the change. Failing instead is the correct outcome, and the collision it guards against is already treated as vanishingly unlikely — the temporary's name carries a millisecond timestamp, and the crate documents that only one `REconverge` writes here.

Leave the rename, the cleanup-on-rename-failure branch, and the error variants as they are.

Add a comment at the site recording why `0o600`: the destination is model data written and read by this service under one account, so nothing else has a reason to open it. That is the decision being recorded, not a claim that the bytes are a secret. Add a second short comment saying what the `sync_all` protects — the classifier file that `load_model_by_name` reads back for a model row already committed to RocksDB.

Add a `### Changed` entry under `## [Unreleased]` in `CHANGELOG.md`. An operator can observe this: a deployment running `0.46.0` under the common `umask 022` has these files at `0o644` today and will have them at `0o600` after the change, so anything reading them as another account stops working. The entry says that the classifier file is now created requesting owner-only permissions instead of the previous `0o666`-minus-umask, and gives the `0o644` to `0o600` shift as a worked example, labelled in the entry itself as what happens with `umask 022`. It must not claim the resulting mode is umask-independent, nor present `0o600` as the mode every deployment will see, because neither is true. The flush is invisible by comparison and gets no entry.

## Acceptance criteria

- [x] `store_classifier` creates its temporary file through `OpenOptions` with `.mode(0o600)` and `.create_new(true)`, with a comment at the site giving the reason above
- [x] The temporary is flushed with `sync_all` before `fs::rename`, with a comment naming what it protects
- [x] A stored classifier file carries no group and no other permission bits — the property that holds under any umask
- [x] Neither the code comments nor `CHANGELOG.md` describe the resulting mode as fixed, guaranteed, or independent of the umask
- [x] The rename, the temp-file cleanup on rename failure, and the `ClassifierFsError` variants are unchanged; no new error variant is added
- [x] `CHANGELOG.md` has the `### Changed` entry described above, under `## [Unreleased]`, with no issue or PR number in it
- [x] `cargo fmt -- --check --config group_imports=StdExternalCrate` and `cargo clippy --bins --tests --all-features -- -D warnings` are clean, and CI is green

## Constraints

- Set the permissions as the file is created and never afterwards. Do not add a `set_permissions` call after the write to force the mode to exactly `0o600` in spite of the umask. `AGENTS.md` rules that out, the umask narrowing the owner bits further is not a problem to solve, and a post-hoc `set_permissions` would override a deliberate operator umask rather than respect it.
- Do it inline. `store_classifier` is the only temp-file-and-rename site in the crate — every other `fs::write` and `fs::rename` under `src/` is in a test — so do not extract a shared atomic-write helper for one caller.
- Keep the existing temporary-path scheme: a sibling of the destination named with a millisecond timestamp extension. It is already in the same directory, which is what makes the rename atomic. Do not switch to `tempfile::NamedTempFile`; `tempfile` is a dev-dependency here and is not available to `src/`.
- Call `.mode(0o600)` without a `cfg` guard. `OpenOptionsExt` is Unix-only, making this the first thing in the crate that will not build on Windows, and that is fine: `src/` has no `cfg(unix)` or `cfg(windows)` anywhere today, and CI builds on `ubuntu-latest` and `macOS-latest` only. Do not add a `cfg` split or a non-Unix fallback for a platform the crate does not claim to support.
- Report a failure to open or write the temporary as the existing `ClassifierFsError::FileWrite(err, temp_path)`, and a failure to `sync_all` the same way. Do not let a `sync_all` error pass silently.
- Add no new dependency.

## Out of scope

- Making the stored file's mode independent of the umask, by any means.
- Syncing the containing directory after the rename. `AGENTS.md` makes that a decision per file rather than a default, because it costs a disk round trip, and nothing in this crate does it today. Flushing the temporary is what closes the "rename visible, contents empty" window; the directory sync only decides whether the rename itself survives.
- The mode of the `classifiers/` and `classifiers/model_{id}` directories, which `fs::create_dir_all` still leaves to the umask, in both `ClassifierFileManager::new` and `store_classifier`.
- The write ordering between the RocksDB model row and the classifier file in `upsert_model`. The crash window it opens is real but is a separate concern from how this one file is written.
- Any change to `load_classifier`, `classifier_exists`, or `delete_classifier`, including making `load_model_by_name` detect a zero-length classifier file.

## Test plan

- [x] A new test in the `tests` module of `src/classifier_fs.rs` stores a classifier under a `tempfile::TempDir` and asserts, via `std::os::unix::fs::PermissionsExt`, that `fs::metadata(path).permissions().mode() & 0o077 == 0` — no group or other access. This is the assertion that is sound under any umask. It catches a regression to `fs::write` in an environment using `umask 022`, where `fs::write` yields `0o644`; under a umask that already strips the group and other bits it cannot tell the two apart, which is the price of an assertion that never fails spuriously.
- [x] The test does not assert exact equality with `0o600`, since a restrictive umask legitimately produces `0o400`
- [x] The test does not set or read the process umask, and does not mutate the process environment
- [x] The existing `test_store_and_load_classifier`, `test_store_creates_parent_directories`, and `test_concurrent_stores` still pass unchanged, confirming the round trip and the rename behaviour are intact
- [x] A second test stores over a destination left group- and other-readable by an earlier release, and asserts the same `& 0o077 == 0` there, that the fresh bytes read back, and that no temporary is left in the model directory. This is the case the permissions argument turns on: the rename installs the temporary's inode, so the destination takes the temporary's mode rather than keeping the one it had
- [x] `cargo test --all-features` passes on Linux and macOS

## Dependencies

None. This issue has no parent, no predecessors, and no successors.

## Pointers

- `src/classifier_fs.rs` — `ClassifierFileManager::store_classifier`, the `fs::write` and `fs::rename` pair; `ClassifierFsError::FileWrite`, `FileRename`, and `TempFileCleanup`; the `#[cfg(test)] mod tests` block that already uses `tempfile::TempDir`
- `src/lib.rs` — `upsert_model`, which commits the model row before calling `store_classifier`, and `load_model_by_name`, which returns `Classifier file not found for model {name}` when the file is absent
- `CHANGELOG.md` — the `## [Unreleased]` section
- `AGENTS.md` — "Files and I/O", for the atomic-write and restrictive-permissions rules

sehkone:
Closes #844

## Summary

`ClassifierFileManager::store_classifier` wrote its sibling temporary through `fs::write` and renamed it into place without flushing. Two consequences, both fixed here:

- **Permissions.** `fs::write` asks for `0o666` and lets the umask subtract. Because `rename` installs the temporary's inode at the destination, the finished classifier ended up with whatever the umask left — commonly `0o644` — and any mode an operator had set by hand was discarded at the next update. The temporary is now created through `OpenOptions` with `.mode(0o600).write(true).create_new(true)`. `create_new` is load-bearing: a requested mode applies only when `open` actually creates the file, so reopening an existing temporary would silently keep its old mode. The mode is set at creation and never afterwards — no `set_permissions` call, so a deployment under a more restrictive umask still narrows its own files further.
- **Durability.** `sync_all` now runs on the temporary before the rename. `upsert_model` commits the model row to RocksDB first and `load_model_by_name` reads the file back for that row, so a rename reaching disk ahead of the data blocks left the database listing a model whose classifier is absent or zero-length — and a truncated file satisfies `classifier_exists` and is returned by `load_classifier` with no error.

Both open/write and `sync_all` failures are reported as the existing `ClassifierFsError::FileWrite(err, temp_path)`. The rename, the cleanup-on-rename-failure branch, and the `ClassifierFsError` variants are unchanged, and no dependency was added. `.mode()` is called without a `cfg` guard, per the issue: `OpenOptionsExt` is Unix-only and CI builds on `ubuntu-latest` and `macOS-latest`.

Comments at the site record why `0o600` (model data this service writes and reads back under one account) and what the flush protects. The permissions comment notes that a umask may narrow the mode further, rather than claiming it is fixed.

`CHANGELOG.md` gains a `### Changed` entry under `## [Unreleased]`, since an operator can observe the shift; it gives `0o644` → `0o600` as a worked example explicitly labelled as `umask 022` and states that the umask still applies. The flush gets no entry.

## Test plan

- [x] New `test_store_grants_no_group_or_other_access` stores under a `TempDir` and asserts `mode() & 0o077 == 0` — sound under any umask
- [x] Neither new test asserts exact equality with `0o600`, since a restrictive umask legitimately yields `0o400`
- [x] Neither new test reads or sets the umask, and neither mutates the process environment
- [x] New `test_store_replaces_a_group_readable_file` seeds a `0o644` destination, stores over it, and asserts `& 0o077 == 0`, that the fresh bytes read back, and that no temporary is left in the model directory
- [x] `test_store_and_load_classifier`, `test_store_creates_parent_directories`, and `test_concurrent_stores` pass unchanged
- [x] `cargo fmt -- --check --config group_imports=StdExternalCrate` clean
- [x] `cargo clippy --bins --tests --all-features -- -D warnings` clean
- [x] `cargo test --all-features` passes locally on macOS (507 passed, 0 failed); Linux covered by CI
- [x] `markdownlint-cli2 CHANGELOG.md` clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Classifier files are now created requesting owner-only permissions (`0o600`)
+  instead of the previous `0o666` reduced by the umask. The umask still applies,
+  so the resulting mode is not fixed, but no group or other bit is ever granted.
+  Under the common `umask 022`, for example, a stored classifier that used to be
+  `0o644` is now `0o600`, so anything reading these files as another account
+  stops working.
 - **BREAKING**: Event timestamps now use `jiff::Timestamp` instead of chrono's
   `DateTime<Utc>`. Existing databases need no migration, because timestamps are
   still stored as `i64` epoch nanoseconds.
--- a/src/classifier_fs.rs
+++ b/src/classifier_fs.rs
@@ -1,5 +1,7 @@
 use std::{
-    fs,
+    fs::{self, OpenOptions},
+    io::Write,
+    os::unix::fs::OpenOptionsExt,
     path::{Path, PathBuf},
 };
 
@@ -81,9 +83,32 @@ impl ClassifierFileManager {
         let timestamp = chrono::Utc::now().timestamp_millis();
         let temp_path = file_path.with_extension(timestamp.to_string());
 
-        // Write to temporary file first
-        fs::write(&temp_path, data)
+        // Write to temporary file first. The mode is requested here because
+        // `rename` installs this inode at the destination, so the finished file
+        // ends up with whatever mode the temporary was created with. `0o600`
+        // records that the destination holds model data this service writes and
+        // reads back under a single account, so nothing else has a reason to
+        // open it; a umask may narrow it further. `create_new` is required for
+        // the mode to take effect, since it is applied only when `open` creates
+        // the file: reopening an existing temporary would keep its old mode.
+        let mut temp_file = OpenOptions::new()
+            .write(true)
+            .create_new(true)
+            .mode(0o600)
+            .open(&temp_path)
             .map_err(|err| ClassifierFsError::FileWrite(err, temp_path.clone()))?;
+        temp_file
+            .write_all(data)
+            .map_err(|err| ClassifierFsError::FileWrite(err, temp_path.clone()))?;
+
+        // Flush before the rename. `upsert_model` has already committed the
+        // model row to RocksDB, and `load_model_by_name` reads this file back
+        // for that row, so a rename reaching disk ahead of the data would leave
+        // the model pointing at an empty or truncated classifier.
+        temp_file
+            .sync_all()
+            .map_err(|err| ClassifierFsError::FileWrite(err, temp_path.clone()))?;
+        drop(temp_file);
 
         // Rename to final location
         if let Err(rename_err) = fs::rename(&temp_path, &file_path) {
@@ -210,7 +235,7 @@ pub enum ClassifierFsError {
 
 #[cfg(test)]
 mod tests {
-    use std::sync::Arc;
+    use std::{os::unix::fs::PermissionsExt, sync::Arc};
 
     use tempfile::TempDir;
 
@@ -271,6 +296,65 @@ mod tests {
 
         let loaded_data = manager.load_classifier(model_id, name).unwrap();
         assert_eq!(loaded_data, test_data);
+    }
+
+    #[test]
+    fn test_store_grants_no_group_or_other_access() {
+        let temp_dir = TempDir::new().unwrap();
+        let manager = ClassifierFileManager::new(temp_dir.path()).unwrap();
+
+        let model_id = 457;
+        let name = "permission_test";
+
+        manager
+            .store_classifier(model_id, name, b"test classifier data")
+            .unwrap();
+
+        let path = manager.create_classifier_path(model_id, name).unwrap();
+        let mode = fs::metadata(&path).unwrap().permissions().mode();
+
+        // Only the absence of group and other bits is asserted: a umask can
+        // clear bits but never add them, so this holds whatever the umask is,
+        // while the exact mode does not.
+        assert_eq!(mode & 0o077, 0, "unexpected mode {mode:o} for {path:?}");
+    }
+
+    #[test]
+    fn test_store_replaces_a_group_readable_file() {
+        let temp_dir = TempDir::new().unwrap();
+        let manager = ClassifierFileManager::new(temp_dir.path()).unwrap();
+
+        let model_id = 458;
+        let name = "replacement_test";
+
+        // Stand in for a classifier left by an earlier release, which wrote
+        // through `fs::write` and so ended up group- and other-readable.
+        let path = manager.create_classifier_path(model_id, name).unwrap();
+        let parent = path.parent().unwrap();
+        fs::create_dir_all(parent).unwrap();
+        fs::write(&path, b"stale data").unwrap();
+        fs::set_permissions(&path, fs::Permissions::from_mode(0o644)).unwrap();
+
+        manager
+            .store_classifier(model_id, name, b"fresh data")
+            .unwrap();
+
+        // The rename installs the temporary's inode, so the destination takes
+        // the temporary's mode rather than keeping the one it had.
+        let mode = fs::metadata(&path).unwrap().permissions().mode();
+        assert_eq!(mode & 0o077, 0, "unexpected mode {mode:o} for {path:?}");
+        assert_eq!(
+            manager.load_classifier(model_id, name).unwrap(),
+            b"fresh data"
+        );
+
+        // The rename consumed the temporary, so the classifier is the only
+        // file left in the model directory.
+        let leftovers: Vec<_> = fs::read_dir(parent)
+            .unwrap()
+            .map(|entry| entry.unwrap().path())
+            .collect();
+        assert_eq!(leftovers, vec![path]);
     }
 
     #[test]


sehkone:
**[Reviewer Round 1]**

No blocking concerns. This implements the intended create-write-sync-rename sequence in `src/classifier_fs.rs`: the exclusive temporary is created requesting `0o600` at [line 94](https://github.com/aicers/review-database/blob/sehkone/issue-844/src/classifier_fs.rs#L94), write and `sync_all` failures retain the existing `FileWrite` mapping ([lines 100-110](https://github.com/aicers/review-database/blob/sehkone/issue-844/src/classifier_fs.rs#L100)), and the existing rename-failure cleanup is left intact. The comments accurately distinguish the no-group/other-bits guarantee from an exact umask-independent mode.

The new tests cover the right externally visible properties: a freshly stored file has no group or other permissions, and replacing a formerly `0o644` file installs fresh data with the restricted mode and leaves no temporary file ([lines 302-360](https://github.com/aicers/review-database/blob/sehkone/issue-844/src/classifier_fs.rs#L302)). The changelog also describes the user-visible permission change without overstating the final mode.

sehkone:

PR #846, opened 2026-08-10
Update shared instruction blocks to 0.2.1

sehkone:
Synced from `aicers/agent-instructions` at 0.2.1.

Blocks in this pull request: workflow agentcoop rust changelog

0.2.0 added the `agentcoop` block — what AgentCoop is, its author and reviewer agents, and what its implementation, design, and verification stages do. That vocabulary is already in this repository's issues and pull request comments, and AgentCoop injects no instructions of its own, so `AGENTS.md` is the only place it can reach a run. 0.2.1 reworded the block's opening sentence; nothing consumed 0.2.0, so this pull request was rebuilt on it rather than opened a second time.

A new block needs its marker pair placed by hand: nothing upstream can decide where in `AGENTS.md` a block belongs, and the apply refuses the repository until the pair is there. The pair sits after the `workflow` block — the system that turns an issue into a pull request reads next to the issue and pull request conventions — and the apply filled it in the same commit. Splitting the two would have left the drift check red in between, since it rejects a marker the pin file does not list.

The marked blocks are generated. Change the wording upstream in `aicers/agent-instructions` rather than editing it here — the drift check in CI will fail if this repository's copy diverges.

### DIFF
--- a/.agents/instructions.toml
+++ b/.agents/instructions.toml
@@ -2,5 +2,5 @@
 # it is on, and the shared blocks it has. Its drift check reads this file,
 # and its apply job rewrites it. Which blocks a repository should carry is
 # decided upstream in repos.json, not here.
-ref = "0.1.5"
-blocks = ["workflow", "rust", "changelog"]
+ref = "0.2.0"
+blocks = ["workflow", "agentcoop", "rust", "changelog"]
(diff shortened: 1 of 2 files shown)


sehkone:
Closed by a branch rename, not on purpose: the head branch `sehkone/instructions-0.2.0` was renamed to `sehkone/instructions-0.2.1` to match the release it now carries, and GitHub treats a rename as deleting the old ref. Same commit, same diff, reopened as #847.

